PR #2, opened 2020-03-18
barebones end-to-end example

dmitryduev:
I got it working with the dockerized version of SkyPortal! However, several things need to be addressed:

- See [this SkyPortal issue](https://github.com/skyportal/skyportal/issues/377)
- See [this SkyPortal issue](https://github.com/skyportal/skyportal/issues/378)
- need to put in a db password into the config file(s). 

With these fixes, running
```bash
./fritz run --init
./fritz run
docker exec -it kowalski_ingester_1 python -m pytest -s test_ingester.py
```
should execute a barebones end-to-end example.

### DIFF
--- a/fritz
+++ b/fritz
@@ -14,6 +14,7 @@ def run(args):
     env.update({"FLAGS": "--config=../fritz.yaml"})
 
     if args.init:
+        # todo: config.yaml.defaults to config.yaml if config.yaml does not exist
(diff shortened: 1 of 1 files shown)


stefanv:
That's fantastic 🚀!

dmitryduev:
Can you please try it out?

PR #4, opened 2020-04-09
Bare-bones end-to-end demo

dmitryduev:
Follow the simple instructions in `readme.md` to run the bare-bones end-to-end demo. For this to work, [this PR on `skyportal`](https://github.com/skyportal/skyportal/pull/400) needs to be addressed and merged first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,5 @@ Initialize `fritz` and run tests:
 ./fritz run --init
 ./fritz test
 ```
+
+Go to `http://localhost:9000/` -- you should see a few real alerts among the displayed sources.
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ Clone the repo from `GitHub`, pull the submodules (`skyportal` and `kowalski`), 
 ```bash
 git clone https://github.com/fritz-marshal/fritz.git && cd fritz
 git submodule update --init
-chmod 755 fritz
 ```
 
 Initialize `fritz` and run tests:
--- a/README.md
+++ b/README.md
@@ -1,3 +1,19 @@
 # fritz
 
 Frontend for Reviewing Interesting Transients from ZTF-II
+
+## Bare-bones end-to-end demo
+
+Clone the repo from `GitHub`, pull the submodules (`skyportal` and `kowalski`), and make the `fritz` utility executable:
+```bash
+git clone https://github.com/fritz-marshal/fritz.git && cd fritz
+git submodule update --init
+chmod 755 fritz
+```
+
+Initialize `fritz` and run tests:
+
+```bash
+./fritz run --init
+./fritz test
+```
--- a/README.md
+++ b/README.md
@@ -4,10 +4,9 @@ Frontend for Reviewing Interesting Transients from ZTF-II
 
 ## Bare-bones end-to-end demo
 
-Clone the repo from `GitHub`, pull the submodules (`skyportal` and `kowalski`), and make the `fritz` utility executable:
+Clone the repo from `GitHub`:
 ```bash
 git clone https://github.com/fritz-marshal/fritz.git && cd fritz
-git submodule update --init
 ```
 
 Initialize `fritz` and run tests:
--- a/fritz
+++ b/fritz
@@ -21,14 +21,17 @@ def status(message):
         print(f'\r[✓] {message}')
 
 
-def check_config(cfgs):
+def check_configs(cfgs):
+    # use config defaults if configs do not exist
     for cfg in cfgs:
-        if not Path(cfg).exists():
-            cd = input(f'{cfg} does not exist, do you want to use {cfg}.defaults (not recommended)? [y/N] ')
+        cwd = str(Path(cfg).parent)
+        c = cfg.replace('.defaults', '')
+        if not Path(c).exists():
+            cd = input(f'{c} does not exist, do you want to use {cfg} (not recommended)? [y/N] ')
             if cd.lower() == 'y':
-                subprocess.run(["cp", f"{cfg}.yaml.defaults", f"{cfg}.yaml"], cwd="skyportal")
+                subprocess.run(["cp", f"{cfg}", f"{c}"], cwd=cwd)
             else:
-                raise IOError(f'{cfg} does not exist, aborting')
+                raise IOError(f'{c} does not exist, aborting')
 
 
 def get_skyportal_token():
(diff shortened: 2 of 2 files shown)


dmitryduev:
Note how ugly token generation for `kowalski` is handled. Wanted to have something that finally works out of the box... Improvement suggestions are most welcome!

dannygoldstein:
@dmitryduev is there any way to do this without using docker? The docker build takes a while to get going. Is it possible to use make / pip to launch the app?

dannygoldstein:
Error when building:

```ERROR: error pulling image configuration: unexpected EOF
Error: No such container: skyportal_web_1
Traceback (most recent call last):
  File "./fritz", line 220, in <module>
    getattr(sys.modules[__name__], args.command)(args)
  File "./fritz", line 112, in run
    update_kowalski_secrets(token)
  File "./fritz", line 72, in update_kowalski_secrets
    raise IOError(f'{p_secrets} does not exist')
OSError: kowalski/secrets.json does not exist```

**Comment by dmitryduev** [comment]
@dannygoldstein Kowalski needs Kafka, which in turn needs Java SDK. You would also need to install a system-wide mongodb and postgres. So, yes, possible, but I don't think we should be doing this, at least for now.

**Comment by dannygoldstein** [comment]
Got it. Makes sense

________________________________

**Comment by dmitryduev** [comment]
@dannygoldstein @stefanv fixed, sorry about that :)

**Comment by dannygoldstein** [comment]
What are the best practices for initializing these files?

```kowalski/secrets.json does not exist, do you want to use kowalski/secrets.defaults.json (not recommended)? [y/N] y
Building Kowalski
Checking configuration
kowalski/config_api.json does not exist, do you want to use kowalski/config_api.defaults.json (not recommended)? [y/N] y
kowalski/config_ingester.json does not exist, do you want to use kowalski/config_ingester.defaults.json (not recommended)? [y/N] y
kowalski/supervisord_api.conf does not exist, do you want to use kowalski/supervisord_api.defaults.conf (not recommended)? [y/N] y
kowalski/supervisord_ingester.conf does not exist, do you want to use kowalski/supervisord_ingester.defaults.conf (not recommended)? [y/N] y
docker-compose.fritz.yaml does not exist, do you want to use docker-compose.fritz.defaults.yaml (not recommended)? [y/N] y```

**Comment by dannygoldstein** [comment]
I confirm that I can run this end-to-end.

File: README.md [hunk 1]
```diff
@@ -1,3 +1,20 @@
 # fritz
 
 Frontend for Reviewing Interesting Transients from ZTF-II
+
+## Bare-bones end-to-end demo
+
+Clone the repo from `GitHub`, pull the submodules (`skyportal` and `kowalski`), and make the `fritz` utility executable:
```
stefanv, line 7: The Fritz utility should check and initialize submodules.

The Fritz utility should also be executable by default in the repository---is it not for you?

(I.e., this step of the instructions, apart from the git clone, should happen implicitly.)

fritz
```diff
@@ -1,9 +1,86 @@
...
-import subprocess
+from contextlib import contextmanager
+import json
 import os
 from pathlib import Path
+import subprocess
+import sys
+
+
+@contextmanager
```
stefanv, line 11: If we start the file with:

```
try:
    import baselayer
except ImportError:
    ... clone baselayer
```

from baselayer import ...

Then we can re-use that functionality instead of duplicating it here.

fritz
```diff
@@ -1,9 +1,86 @@
...
+    try:
+        yield
+    except:
+        print(f'\r[✗] {message}')
+        raise
+    else:
+        print(f'\r[✓] {message}')
+
+
+def check_configs(cfgs):
```
stefanv, line 24: We can just make empty configuration files if they don't exist, then they won't override the defaults.

fritz
```diff
@@ -1,9 +1,86 @@
...
+        c = cfg.replace('.defaults', '')
+        if not Path(c).exists():
+            cd = input(f'{c} does not exist, do you want to use {cfg} (not recommended)? [y/N] ')
+            if cd.lower() == 'y':
+                subprocess.run(["cp", f"{cfg}", f"{c}"])
+            else:
+                raise IOError(f'{c} does not exist, aborting')
+
+
+def get_skyportal_token():
```
stefanv, line 37: Fine for now; any reason not to use an API call instead?

stefanv: Blind moment, just ignore.

dmitryduev:
> What are the best practices for initializing these files?
> 
> ```json
> Building Kowalski
> Checking configuration
> kowalski/config_api.json does not exist, do you want to use kowalski/config_api.defaults.json (not recommended)? [y/N] y
> kowalski/config_ingester.json does not exist, do you want to use kowalski/config_ingester.defaults.json (not recommended)? [y/N] y
> kowalski/supervisord_api.conf does not exist, do you want to use kowalski/supervisord_api.defaults.conf (not recommended)? [y/N] y
> kowalski/supervisord_ingester.conf does not exist, do you want to use kowalski/supervisord_ingester.defaults.conf (not recommended)? [y/N] y
> docker-compose.fritz.yaml does not exist, do you want to use docker-compose.fritz.defaults.yaml (not recommended)? [y/N] y```
> ```

All the practices I know of I don't like, this is why I did it this way. People are usually simply instructed to copy and edit the defaults manually. @stefanv are there better ways?

README.md
dmitryduev: Added submodule init, see [here](https://github.com/fritz-marshal/fritz/pull/4/commits/aa65efb83003953a612fdcb11bc985c4637a3365)

The utility is executable, fixed the instructions.

fritz
dmitryduev: I agree with your point. I just wanted to minimize the number of dependencies for the Fritz utility itself (it's now only using the "stdlib" stuff), so just copied it over. We can actually remove that function for now, it's not used anyways.

fritz
dmitryduev: Kowalski would be expecting values for some settings, so empty ones would raise errors. I think SkyPortal also didn't like empty db pwds or something. Can you elaborate on your point, I may be misunderstanding something?

fritz
dmitryduev: :) Ari helped me with this, so this is what we arrived at after a couple iterations.

dmitryduev:
@dannygoldstein @stefanv Can we merge now?

stefanv:
Thanks @dmitryduev !

PR #8, opened 2020-05-19
Single config file for Kowalski and SkyPortal

dmitryduev:
- Kowalski updated to use a single config yaml file
- Kowalski's and SkyPortal's default `config.yaml` files merged into `fritz.defaults.yaml` (renamed `fritz.yaml`)
- `fritz.yaml` is copied to `kowalski` and `skyportal`
- At `./fritz run --init`, a Kowalski token is generated and written to SkyPortal's config file
- Once SkyPortal is up and running, a token is generated for Kowalski and written into its config file
- Added prune command to the fritz utility
- Reflect the `Obj` stuff in the API calls to SkyPortal in 'test_ingester.py'


dmitryduev:
@stefanv filed a skyportal [issue](https://github.com/skyportal/skyportal/issues/450) discovered  while testing this PR -- thumbnails do not persist when running with docker-compose

dmitryduev:
@stefanv Fritz can now generate a Kowalski token for Skyportal and write it into the latter's config.yaml, but the docker build uses docker.yaml (hardcoded in Dockerfile). What would be a good way to overcome this? [How about renaming docker.yaml into docker.yaml.defaults? Don't want to write stuff to docker.yaml]

stefanv:
@dmitryduev The `docker.yaml` file is used when the Dockerfile is built, but you can also mount a file on top of `config.yaml` during deployment.  Would that solve your problem?

stefanv:
@dmitryduev I updated the CI script to install the fritz dependencies first, so hopefully your next commit will build successfully.

fritz
```diff
@@ -21,24 +23,62 @@ def status(message):
...
-                    raise IOError(f"Failed to copy {cfg} to {c}")
-            else:
-                raise IOError(f'{c} does not exist, aborting')
+def check_config(cfg='fritz.defaults.yaml'):
+    c = cfg.replace('.defaults', '')
+    if not Path(c).exists():
+        cd = input(f'{c} does not exist, do you want to use {cfg} (not recommended)? [y/N] ')
+        if cd.lower() == 'y':
+            s = subprocess.run(["cp", f"{cfg}", f"{c}"], check=True)
+            if s.returncode != 0:
```
stefanv, line 32: If you have `check=True`, then you don't need to check the returncode.  I think `check=True` is probably the right way to go for most of these cases, so we don't have to handle the returncodes each time.

fritz
```diff
@@ -21,24 +23,62 @@ def status(message):
...
+                raise IOError(f"Failed to copy {cfg} to {c}")
+        else:
+            raise IOError(f'{c} does not exist, aborting')
+
+    # generate a token for SkyPortal to talk to Kowalski:
+    with open(c) as cyaml:
+        config = yaml.load(cyaml, Loader=yaml.FullLoader)
+
+    kowalski_token = generate_kowalski_token(
```
stefanv, line 41: Ideally move the function definition earlier on (forward references can be quite confusing while reading the code).

fritz
```diff
@@ -21,24 +23,62 @@ def status(message):
...
+        user_id=config['kowalski']['server']['admin_username'],
+        jwt_secret=config['kowalski']['server']['JWT_SECRET_KEY'],
+    )
+    config['app']['kowalski_token'] = kowalski_token
+
+    with open(c, 'w') as cyaml:
+        yaml.dump(config, cyaml)  # default_flow_style=False
```
stefanv, line 48: What does the comment mean here?

fritz
```diff
@@ -172,6 +200,43 @@ def lint(args):
     subprocess.run(["pre-commit", "run", "--all-files"])
 
 
+def prune(args):
+    """
+    Prune fritz'es docker containers and volumes and reset configs to defaults
```
stefanv, line 205:
```suggestion
    Prune fritz's docker containers and volumes and reset configs to defaults
```

fritz
```diff
@@ -172,6 +200,43 @@ def lint(args):
...
+    """
+    go = input(f"Do you want to prune fritz'es docker containers and volumes and deinit submodules? [y/N] ")
```
stefanv, line 207: Maybe we should provide a generic `--yes` option for all commands so that they can be re-used inside of CI where we have no interactive input.  Happy to add that myself later.

stefanv, line 207:
```suggestion
    go = input(f"Do you want to prune fritz's docker containers and volumes and deinit submodules? [y/N] ")
```

fritz
```diff
@@ -172,6 +200,43 @@ def lint(args):
...
+
+    if go.lower() == 'y':
+        # try stopping anything that's running first:
+        stop(args)
+
+        # remove docker images
+        for image_name in ('kowalski_api', 'kowalski_ingester', 'skyportal/web'):
+            p1 = subprocess.Popen(['docker', 'images'], stdout=subprocess.PIPE)
+            p2 = subprocess.Popen(['grep', image_name], stdin=p1.stdout, stdout=subprocess.PIPE)
+            image_id = subprocess.check_output(['awk', '{print $3}'], stdin=p2.stdout, universal_newlines=True).strip()
```
stefanv, line 217: Heh, I read this and thought: "isn't this why we have Python?"  But, looking at the Python version I think the Unix pipes version actually reads more clearly :man_shrugging: 

```python
image_lines = subprocess.check_output([
    'docker', 'images',
    '--format', '{{.Repository}} {{.ID}}'
]).decode('utf-8').strip().split('\n')

images = [x.split() for x in image_lines]
removed_ids = [image_id for (repo, image_id) in images if image_name in repo_id]
```

fritz
```diff
@@ -172,6 +200,43 @@ def lint(args):
...
+            p = subprocess.run(['docker', 'volume', 'rm', volume_name])
+            if p.returncode == 0:
+                print(f'Removed {volume_name} docker volume')
+            else:
+                print(f'Failed to remove {volume_name} docker volume')
+
+        # deinit submodules
+        p = subprocess.run(['git', 'submodule', 'deinit', '--all', '-f'])
+        if p.returncode == 0:
+            print(f"Deinitialized fritz' submodules")
```
stefanv, line 235:
```suggestion
            print(f"Deinitialized fritz's submodules")
```

fritz
```diff
@@ -172,6 +200,43 @@ def lint(args):
...
+        else:
+            print(f"Failed to deinit fritz' submodules")
```
stefanv, line 237:
```suggestion
            print(f"Failed to deinit fritz's submodules")
```

README.md
```diff
@@ -9,6 +9,12 @@ Clone the repo from `GitHub`:
 git clone https://github.com/fritz-marshal/fritz.git && cd fritz
 ```
 
+Make sure the requirements to run the `fritz` utility are met, e.g.:
+
+```bash
+pip install -r requirements.txt --no-cache-dir
```
stefanv, line 15:
```suggestion
pip install -r .requirements/dev.txt
```

stefanv:
@dmitryduev Let's also try and make `fritz` work as much as it can without trying to import all dependencies.  This would be helpful in CI, so that I can build the docs without installing all dev dependencies, e.g.

The way to fix that here is to import `jwt`, etc. in the function where it is used.

stefanv:
(@dmitryduev I'm happy to make these suggestions as a PR, if that'd be helpful.)

fritz
dmitryduev: Removed the redundant checks.

fritz
dmitryduev: Done.

fritz
dmitryduev: Nothing, just forgot to remove it - was experimenting to see which style I like more. Removed.

fritz
dmitryduev: Exactly. I thought that since we have to use subprocess to ask docker anyways, so why not just do the whole thing with the Unix pipes.

dmitryduev:
@stefanv Implemented all the suggestions -- can we merge now?

stefanv:
Thanks @dmitryduev.  I'm back-pedaling on my suggestion to have to import libraries only where they're used; it complicates things too much.  We'll just make sure users first do `pip install -r requirements.txt` (pretty standard in Python projects), and then they're good to go.

dmitryduev:
Cool, thanks @stefanv!

PR #9, opened 2020-06-05
Update demo

dmitryduev:
- Update K and SP
  - End-to-end demo (`./fritz test`) now posts to /candidates
- Add two alert filter examples
- Update `fritz.defaults.yaml` with K data for SP


dmitryduev:
@stefanv Not much to review here, this is mostly to update K and SP

PR #11, opened 2020-06-30
Alert page

dmitryduev:
This adds the ZTF Alert page that displays the detection history, photometry plot and cutouts for a ZTF `objectId`.  

The PR introduces:
- Submodule extensions such as Fritz-specific SP API handlers and endpoints
- Fritz-specific SP API endpoints to get alerts, auxiliary data (`prv_candidates`, `cross_matches` etc.), and cutouts for a given `objectId` from K
- Dependency/env check for the `fritz` utility
- Fritz-specific `docker-compose.skyportal.yaml` file
- `app_server_fritz.py` file that overloads SP's default app factory
- The ZTF alert page
  - Vega-lite plot to display photometry (5-sigma alerts + 3-sigma detections + upper-limits for non-detections) on the ZTF Alert page
  - React table to display 5-sigma alerts with pagination and sortable columns
  - Dynamically loaded cutouts for a selected alert
- Alerts page to search alert streams for objects
- Different/matching color for the SP's top navbar :)
- K and SP updates to latest versions

The code currently lacks tests (Fritz-specific, of the corresponding SP/K extensions), will use help from @dannygoldstein to set these up.

ztf-alert-01

image


dannygoldstein:
This is cool 'n' awesome

dannygoldstein:
I was able to get the tests to pass locally. How come no CI?

dannygoldstein:
@dmitryduev  Merge conflicts

extensions/skyportal/skyportal/app_server_fritz.py
```diff
@@ -0,0 +1,40 @@
+from skyportal.app_server import make_app
+
+from skyportal.handlers.api.alert import (
```
dannygoldstein, line 3: Let's see, so these `skyportal`s are being imported from two different places, right? `make_app` is coming from the submodule, but `handlers.api.alert` is coming from the extension. What assumptions have to be made about the user path for this to work?

extensions/skyportal/static/js/components/VegaPlotZTFAlert.jsx
```diff
@@ -0,0 +1,244 @@
...
+          }
+        }
+    />
+);
+
+VegaPlot.propTypes = {
+  dataUrl: PropTypes.string.isRequired
+};
+
+export default VegaPlot;
```
dannygoldstein, line 244: A lot of this file seems to be present in `VegaPlot.jsx`, can any of the code there be recycled?

extensions/skyportal/static/js/ducks/alert.js
```diff
@@ -0,0 +1,84 @@
...
+export const FETCH_AUX_ERROR = 'skyportal/FETCH_AUX_ERROR';
+export const FETCH_AUX_FAIL = 'skyportal/FETCH_AUX_FAIL';
+
+// export const FETCH_CUTOUT = 'skyportal/FETCH_CUTOUT';
+// export const FETCH_CUTOUT_OK = 'skyportal/FETCH_CUTOUT_OK';
+// export const FETCH_CUTOUT_ERROR = 'skyportal/FETCH_CUTOUT_ERROR';
+// export const FETCH_CUTOUT_FAIL = 'skyportal/FETCH_CUTOUT_FAIL';
+
+export function fetchAlertData(id) {
+    return API.GET(`/api/alerts/ztf/${id}`, FETCH_ALERT)
```
dannygoldstein, line 21: Is there a way to implement this without having `ztf` hardcoded in? Otherwise, we will have to reimplement separate action dispatch functions for Gattini, LSST, etc.

fritz
```diff
@@ -57,7 +47,7 @@ def generate_skyportal_token():
          """from skyportal.models import init_db; """
          """from skyportal.model_util import create_token; """
          """conn = init_db(**load_config(['config.yaml'])['database']); """
-         """print(create_token(['Upload data', 'Manage sources'], 1, 'kowalski'))" """
+         """print(create_token(['Upload data', 'Manage sources', 'Manage groups'], 1, 'kowalski'))" """
```
dannygoldstein, line 50: @stefanv, can your new token stuff be used to deprecate this?

fritz
```diff
@@ -72,6 +62,9 @@ def generate_skyportal_token():
 
 
 def check_config(cfg='fritz.defaults.yaml', yes=False):
+    """
```
dannygoldstein, line 65: Do these config files get left behind outside the container after a user stops a container? Might be good to have them generated in a temporary directory and the copied into the container so that the default config file always remains unchanged

fritz
```diff
@@ -123,16 +129,24 @@ def run(args):
...
+
         # Build skyportal's images
         p = subprocess.run(["make", "docker-local"], cwd="skyportal")
         if p.returncode != 0:
             raise RuntimeError("Failed to build skyportal's docker images")
 
     # start up skyportal
-    p = subprocess.run(["docker-compose", "up", "-d"], cwd="skyportal")
+    # create kowalski_kowalski network if it does not exist yet
+    p = subprocess.run(["docker", "network", 'create', "kowalski_kowalski"])
```
dannygoldstein, line 147: Re: our conversation on slack, could be good to append a SHA to `kowalski_kowalski` to avoid collisions

fritz.defaults.yaml
```diff
@@ -17,18 +17,34 @@ app:
...
-
-  kowalski_token:
+    - path: "/upload_photometry/:id"
+      component: UploadPhotometry
+    - path: "/about"
+      component: About
+#    - path: "/alerts"
+#      component: Alerts
+    - path: "/alerts/ztf/:id"
+      component: ZTFAlert
```
dannygoldstein, line 41: Again, is it necessary that `Alert`s from different instruments live in different tables and have different endpoints? Since `Alert`s are unstructured JSON, I think it would be fine to have them all live in a single table with an index on `instrument_id`.

tools/check_environment.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
```
dannygoldstein, line 1: Can this and the other tool be imported from `baselayer`? @stefanv

dannygoldstein:
@dmitryduev In https://github.com/dmitryduev/fritz/tree/alert_page/extensions/skyportal/skyportal/tests/api, we should set up a `conftest.py` file to start building out K fixtures.

extensions/skyportal/skyportal/app_server_fritz.py
dmitryduev: No, the `fritz` script first copies over the extensions to the submodule directories - see [here](https://github.com/fritz-marshal/fritz/blob/76b8429a277a555e05a3b9917caa822191ab8254/fritz#L135) so the imports here are from the same place.

extensions/skyportal/static/js/components/VegaPlotZTFAlert.jsx
dmitryduev: Well, it's the specs (think config essentially) so they do look very similar. This is one is specific to what the alert packets look like and `VegaPlot.jsx` - to how a saved Obj photometry looks like, so not sure whether we need to recycle the code here. @stefanv what do you think?

extensions/skyportal/static/js/ducks/alert.js
dmitryduev: Yea, I know. I actually first implemented this without hardcoding ZTF and then re-did it this way since I thought people might appreciate the "explicitness". 
What would be the alternative? Resolving the alert stream by ${id} (e.g. startswith "ZTF")?

fritz
dmitryduev: I would hope so, but it needs to be merged first -- aren't there any clashes with Josh's stuff (Classification #2)?

fritz
dmitryduev: The default config stays untouched. When you do `fritz run --init`, if you don't have `fritz.yaml`, `fritz` will ask if you want to `cp fritz.defaults.yaml fritz.yaml` warning that it is not recommended. `fritz.yaml` then gets dissected and distributed to K and SP ([here](https://github.com/fritz-marshal/fritz/blob/76b8429a277a555e05a3b9917caa822191ab8254/fritz#L91)).

fritz
dmitryduev: Let's do this in a separate PR? Note for our future selves: the SHA should be set in `fritz.yaml` and propagated to the appropriate places. Filed an issue #14.

fritz.defaults.yaml
dmitryduev: They don't live in SP tables -- SP just makes K API calls and renders the responses. Alerts would come from different collections in K's mongo so the calls would be different. But it's fine, we can figure out what to query by the object name (see above) -- will implement this.

tools/check_environment.py
dmitryduev: Sure, I just want to keep the deps of `fritz` at the very minimum, but don't really care as it's already using pyjwt and pyyaml (I originally wanted stdlib stuff only)

dmitryduev:
> @dmitryduev In https://github.com/dmitryduev/fritz/tree/alert_page/extensions/skyportal/skyportal/tests/api, we should set up a `conftest.py` file to start building out K fixtures.

It seems general enough to go into a separate PR, don't you think @dannygoldstein? Or should we do it here?

fritz.defaults.yaml
dmitryduev: I would actually prefer keeping this as is: the logic for rendering alerts from different streams will likely differ significantly, so I don't want to dump everything into a single component.

dmitryduev:
Added a simplistic `/alerts` page
image

dmitryduev:
@dannygoldstein given my replies to your comments, shall we merge this and implement CI in the next PR?

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -0,0 +1,139 @@
...
+    };
+
+
+    return (
+        <div>
+            <h2>
+                Search objects from alert streams
+            </h2>
+
+            <Grid container spacing={2}>
```
stefanv, line 93: We've switched to hook form for skyportal: https://react-hook-form.com/api/

This is fine, but take a look---it's quite nice.

extensions/skyportal/static/js/components/ProfileDropdown.css
```diff
@@ -0,0 +1,85 @@
+.profileDesktop {
```
stefanv, line 1: Why is this file duplicated?

extensions/skyportal/static/js/components/SidebarAndHeader.jsx
```diff
@@ -0,0 +1,186 @@
+import React from "react";
```
stefanv, line 1: This should not be included here.

extensions/skyportal/static/js/components/VegaPlotZTFAlert.jsx
```diff
@@ -0,0 +1,244 @@
...
+          datum: jd,
+          type: "quantitative"
+        }
+      }
+    }
+  ]
+});
+
+
+const VegaPlot = ({ dataUrl, jd }) => (
```
stefanv, line 228: Should probably have another name?

stefanv: In fact, can't we reuse VegaPlot from SkyPortal and just provide a different schema?

extensions/skyportal/static/js/components/ZTFAlert.jsx
```diff
@@ -0,0 +1,476 @@
...
+
+function EnhancedTableHead(props) {
+    const {classes, order, orderBy, onRequestSort} = props;
+    const createSortHandler = (property) => (event) => {
+        onRequestSort(event, property);
+    };
+
+    return (
+        <TableHead>
+            <TableRow>
```
stefanv, line 217: Are these indents of size 3?  I think we use 2.

tools/check_environment.py
stefanv: baselayer is already there via skyportal submodule, so yes.

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -0,0 +1,433 @@
...
+                stamp_fits = io.BytesIO()
+                hdul.writeto(fileobj=stamp_fits)
+
+                self.set_header("Content-Type", 'image/fits')
+                self.set_header("Content-Disposition", f'Attachment;filename={fits_name}')
+                self.write(stamp_fits.getvalue())
+
+            if file_format == 'png':
+                buff = io.BytesIO()
+                plt.close('all')
```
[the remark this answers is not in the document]
dmitryduev: Fixed.

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -0,0 +1,433 @@
+from astropy.io import fits
+from bson.json_util import loads, dumps
+import gzip
+import io
+from matplotlib.colors import LogNorm
+import matplotlib.pyplot as plt
+from numpy import array, flipud, median, nan_to_num
```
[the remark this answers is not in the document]
dmitryduev: Fixed the imports. Honestly, no idea why I was doing `from numpy import ...` here, never do it like that.

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -0,0 +1,433 @@
...
+import os
+# import pandas as pd
```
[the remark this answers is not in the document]
dmitryduev: Removed all the commented code.

extensions/skyportal/static/js/components/ProfileDropdown.css
dmitryduev: The background color! It's very subjective, but I think that SP's corporate light blue color does not match the side bar and everyone liked Josh's original Figma template, including the rough color scheme. So that's my way of overriding the default colors :) The same goes for `SidebarAndHeader.jsx`. 
What is the proper way of overriding/styling?

extensions/skyportal/static/js/components/SidebarAndHeader.jsx
dmitryduev: Well, I feel the same way, but, omitting the colors, where/how do we define Fritz-specific links in the sidebar - see the snapshot [above](https://github.com/fritz-marshal/fritz/pull/11#issuecomment-659217519)? Sources and Groups should probably go to SP, but Alerts are F-specific.

extensions/skyportal/static/js/components/ZTFAlert.jsx
dmitryduev: My pycharm was configured to use 4 for js code for some reason... Changed to 2

tools/check_environment.py
dmitryduev: In baselayer, it's a stand-alone script that is used in `make dependencies`, and I wanted an importable function. Plus, it would say things like "Please refer to https://cesium-ml.org/baselayer for installation instructions." :)

extensions/skyportal/static/js/components/ProfileDropdown.css
stefanv: You're right---there's no straightforward way to override that.  Unfortunately, overwriting the CSS file may well stop working at any point when we switched to styled components or similar.

Why don't we just make this the default in SkyPortal for now; I don't mind.

extensions/skyportal/static/js/components/SidebarAndHeader.jsx
stefanv: The JS components that appear in the sidebar are supposed to be configurable in the config file.  I guess I thought we had already made that change!  If not, let me do that for you now.

stefanv:
I just saw your comment on the baselayer script not being part of baselayer, but living in the tools directory.  Fair point!  I can refactor that for you at some point, but duplicating the script is fine for now.

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -0,0 +1,139 @@
...
+                color="primary"
+                className={classes.search_button}
+            >
+                Search
+            </Button>
+
+        </div>
+    )
+};
+
```
acrellin, line 138: Add `propTypes` here for `route` prop

extensions/skyportal/static/js/components/VegaPlotZTFAlert.jsx
acrellin: Agreed

extensions/skyportal/static/js/components/ZTFAlert.jsx
```diff
@@ -0,0 +1,476 @@
...
+    padding: theme.spacing(2),
+    textAlign: 'center',
+    color: theme.palette.text.secondary,
+  },
+// table: {
+//     minWidth: 650,
+// },
+}));
+
+function createRows(id, jd, fid, mag, emag, rb, drb, isdiffpos, programid, alert_actions) {
```
acrellin, line 90: this just bundles the arguments into an object? is this worth defining as a function rather than doing in-line?

extensions/skyportal/static/js/components/SidebarAndHeader.jsx
dmitryduev: Yes, please :)

extensions/skyportal/static/js/components/VegaPlotZTFAlert.jsx
dmitryduev: Please teach me the ways to do that properly then, masters :) Specifically, I need an extra parameter passed to VegaPlot (jd) -- how do I reuse the one from SP?

dannygoldstein:
Massive one, GJ @dmitryduev !!

extensions/skyportal/static/js/components/Alerts.jsx
dmitryduev: Will redo with hook form in a future PR

PR #20, opened 2020-08-01
WIP: Alert Streams and Filters #1

dmitryduev:
This PR introduces the infrastructure for managing alert streams and filters in Fritz.

Concretely:

- `extensions/skyportal/configs/db_fritz.yaml` is a first draft of the the seed config file to be used when initializing Fritz
  - Alert streams are defined in the config
- Fritz-specific K & SP data model adjustments + API endpoints to deal with alert streams and filters. The respective concerns of K and SP are separated as much as possible.
- Reworked and expanded Group management page
- Filter management page
- A couple MUI-re-skinned components here and there :)

streams01

The scree capture ⬆ is slightly outdated -- now all successful actions are accompanied by notifications.


dmitryduev:
TODO: 
- changes in `data_loader.py` + `db_firtz.yaml` should take into account @stefanv's latest PR
- several test are missing
- test filter button doesn't do anything currently

But otherwise should be ready to be looked at, @stefanv @dannygoldstein -- mostly to check if there's anything glaring that needs to be fixed straight away.

profjsb:
Small UI thing: I think fields that say "email" or "email addresses" should be changed to "user names" or "usernames". We've been using username == email address in Skyportal as the default but this is not enforced anywhere and, actually, I prefer usernames to be potentially distinct from emails.

extensions/skyportal/static/js/components/Group.jsx
```diff
@@ -0,0 +1,568 @@
+import React, {useEffect, useState} from 'react';
```
acrellin, line 1: @dmitryduev could these extensions/duplications be obviated if we simply added a few features/changes on the SP side?

dmitryduev: that's the hope :) I'm happy to submit the corresponding PRs on the SP side once we converge on what could be "transferred"

dmitryduev:
Closing in favor of #21

PR #24, opened 2020-08-31
Remove unused ZTFAlert.css file

kmshin1397:
As far as I can tell, we're not importing or using the styles in ZTFAlert.css anywhere. Safe to remove?


dmitryduev:
👍

PR #50, opened 2020-09-30
update SP and K

dmitryduev:


dmitryduev (commented):
Just pulling in the latest SP and K

PR #60, opened 2020-10-02
add git logging to the SP data dir

profjsb:
Build build, get the git log and send it into SP so it can be used therein.

### DIFF
--- a/fritz
+++ b/fritz
@@ -177,7 +177,7 @@ def build(args):
         ["git", "--git-dir=.git", "log",
          "--pretty=format:'%C(cyan)[%ci]%Creset %s %C(auto)%h'"],
         capture_output=True,
-        universal_newlines=True, 
+        universal_newlines=True,
         cwd="skyportal"
     )
     out = p.stdout
--- a/fritz
+++ b/fritz
@@ -173,13 +173,14 @@ def build(args):
         f.write(''.join(out).encode('utf-8'))
 
     # get the git log for SP and add it to the SP distro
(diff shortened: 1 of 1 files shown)


fritz
```diff
@@ -172,6 +172,17 @@ def build(args):
...
         f.write(''.join(out).encode('utf-8'))
 
+    # get the git log for SP and add it to the SP distro
+    p = subprocess.Popen(
+        ["git", "--git-dir=skyportal/.git", "log",
+         "--pretty=format:'%C(cyan)[%ci]%Creset %s %C(auto)%h'"],
+        stdout=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    out, err = p.communicate()
```
dmitryduev, line 182: Since no further piping is required here, maybe use `subprocess.run`?

```suggestion
    p = subprocess.run(
        ["git", "--git-dir=.git", "log",
         "--pretty=format:'%C(cyan)[%ci]%Creset %s %C(auto)%h'"],
         capture_output=True,
         universal_newlines=True, 
         cwd="skyportal"
     )
     out = p.stdout
```

dmitryduev (approved):

dmitryduev:
Thanks @profjsb. Weird, linting failed. Could you please `./fritz lint` and commit the result?

stefanv:
If we write this out as a json file to the static directory somewhere we can import it directly from a component.

profjsb:
Yes that is what we do in SP.

PR #64, opened 2020-10-08
Add GROWTH Marshal logo and update SP

kmshin1397:
Adding the actual image file and configuration for the Fritz side of [SP#1057](https://github.com/skyportal/skyportal/pull/1057)

I think I also found a bug in ./fritz with the missing newline.


dmitryduev (approved):
Looks great, thanks @kmshin1397 !

PR #65, opened 2020-10-08
Update git log format to match what SkyPortal expects

stefanv:


dmitryduev (approved):
Looks good, thanks @stefanv!

PR #66, opened 2020-10-09
Switch back to synchronous Kowalski calls

dmitryduev:
SQLAlchemy was sometimes unhappy about the async calls I put in -- would drop the db connection prematurely. Should investigate this at some point, but switching to sync calls for now.


stefanv (approved):

kmshin1397 (approved):
@dmitryduev Merge conflict in SP but otherwise LGTM

PR #75, opened 2020-10-24
Allow saving ZTF alerts to any group, respecting stream access

dmitryduev:
In this PR:

- Allow saving ZTF alerts to any group using `/api/alerts/ztf/<objectId>.POST`. Group (ZTF alert) stream access is respected when saving the photometry.
- Make the `SAVE` button in the modal inactive while the submission is ongoing
- Pin latest SP and K
- Update default config


stefanv (approved):
@acrellin Can you take a look as well?

kmshin1397 (approved):

PR #80, opened 2020-11-03
TNS cross-match data

dmitryduev:
In this PR:

- In `/api/alerts/ztf/<oid>/aux`, run a cone search on the TNS collection on Kowalski and report the results as a part of the `cross_matches` field.

tns-x-match


dannygoldstein (changes requested):
Can we do this through the new fritz extension strategy?

dmitryduev:
@dannygoldstein I'd merge this in and do the migration to the new strategy in the next PR, please.

kmshin1397 (approved):
LGTM! One quick question - why did you remove the `async` on the GET handler? I don't think I saw anything that required making it synchronous explicitly but that may just be me not being as familiar with K

dannygoldstein (approved):

dmitryduev:
@kmshin1397 re your question: it's actually `self.finish()` that would require awaiting should I keep the get def async, so since nothing is really async in this method, I decided to remove it for now.

PR #120, opened 2020-12-03
Customize title and URL of API docs

stefanv:
This relies on https://github.com/skyportal/skyportal/pull/1456 being merged and pinned first.

Update: that has now been merged. This should be good to go.

### DIFF
--- a/fritz
+++ b/fritz
@@ -182,19 +182,28 @@ def patch_skyportal():
 
     # add Fritz-specific dependencies for SP
     # js
-    with open("extensions/skyportal/package.fritz.json", "r") as fpjf, open("skyportal/package.json", "r") as fpj:
-        pjf = json.load(fpjf)
-        pj = json.load(fpj)
-    pj["dependencies"] = {**pj["dependencies"], **pjf["dependencies"]}
-    with open("skyportal/package.json", "w") as fpj:
-        json.dump(pj, fpj, indent=2)
+    with open("extensions/skyportal/package.fritz.json", "r") as f:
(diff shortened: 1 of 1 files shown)


fritz
```diff
@@ -182,19 +182,28 @@ def patch_skyportal():
...
-            fr.write(line)
+    with open(".requirements/ext.txt", "r") as f:
+        ext_req = f.readlines()
+    with open("skyportal/requirements.txt", "r") as f:
+        skyportal_req = f.readlines()
+    with open("skyportal/requirements.txt", "w") as f:
+        f.writelines(skyportal_req)
+        for line in ext_req:
+            if line not in skyportal_req:
+                f.write(line)
```
kmshin1397, line 206: Not sure if this will break anything, but what would happen if `ext_req` and `skyportal_req` had lines requiring the same package but different versions, for example? That is, if a package appears twice in the requirements.txt with different versions tagged how does pip resolve that? Do you know?

stefanv: pip will probably get very unhappy about that.

stefanv: (With the new resolver that they just released.)

stefanv: If we want to override version, we'd have to first parse the requirements and compare.  That's outside the scope of this PR, but something we could potentially add.

kmshin1397: Cool - and that will in turn get propagated as an error at some step when trying to stand up Fritz for the dev to resolve, right?

kmshin1397: Right yeah I think this is fine for now

kmshin1397 (approved):

PR #166, opened 2021-02-23
Fix configuration bugs

kmshin1397:
Two errors I ran into when spinning up Fritz locally are addressed here:
- The Kowalski token was being written to the configs encoded as binary (with `!!binary`) and that didn't agree with SP.
- The `homepage_grid` parameter was missing for the default app configs, leading to an error when trying to build the homepage from its template file. 

The rest of the changes are just from black, I think - not sure why they weren't changed before.

### DIFF
--- a/fritz
+++ b/fritz
@@ -110,7 +110,7 @@ def check_config(cfg="fritz.defaults.yaml", yes=False, **kwargs):
             config["kowalski"]["server"]["JWT_EXP_DELTA_SECONDS"]
         ),
     )
-    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token
+    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token.decode("utf-8")
 
     # get Docker gateway IP
     command = [
--- a/fritz
+++ b/fritz
@@ -110,7 +110,8 @@ def check_config(cfg="fritz.defaults.yaml", yes=False, **kwargs):
             config["kowalski"]["server"]["JWT_EXP_DELTA_SECONDS"]
         ),
     )
-    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token.decode("utf-8")
+
+    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token
 
     # get Docker gateway IP
     command = [
--- a/fritz.defaults.yaml
+++ b/fritz.defaults.yaml
@@ -112,6 +112,40 @@ skyportal:
         icon: Info
         url: /about
 
+  homepage_grid:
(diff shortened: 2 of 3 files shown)


kmshin1397:
Hm looks like CI actually doesn't have the issue I was having locally - investigating

fritz
```diff
@@ -110,7 +110,7 @@ def check_config(cfg="fritz.defaults.yaml", yes=False, **kwargs):
             config["kowalski"]["server"]["JWT_EXP_DELTA_SECONDS"]
         ),
     )
-    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token
+    config["skyportal"]["app"]["kowalski"]["token"] = kowalski_token.decode("utf-8")
```
dmitryduev, line 113: Ah, I've finally realized what's going on: please revert this change and instead bump the required version for `pyjwt>=2.0` in fritz/.requirements/dev.txt. You probably have 1.7.x installed locally -- they made this change in 2.0. The nice thing is that the CI actually caught this!

fritz.defaults.yaml
```diff
@@ -112,6 +112,40 @@ skyportal:
         icon: Info
         url: /about
 
+  homepage_grid:
```
dmitryduev, line 115: Nice catch! I wonder if we have it in prod...

fritz.defaults.yaml
```diff
@@ -143,11 +177,11 @@ skyportal:
     WeatherWidget:
       resizeable: false
       layouts:
-        xlg: [ 0, 10, 4, 1 ]
-        lg: [ 0, 10, 4, 1 ]
-        md: [ 0, 10, 4, 1 ]
-        sm: [ 0, 10, 4, 1 ]
-        xs: [ 0, 12, 4, 1 ]
+        xlg: [0, 10, 4, 1]
```
dmitryduev, line 180: @acrellin was also reporting weird linting behavior. Might be worth looking into why this is happening.

dmitryduev (changes requested):
👍🏻

fritz
kmshin1397: Yup I just came to the same conclusion. Reverted and requirements have been bumped

fritz.defaults.yaml
kmshin1397: Yea I ran `./fritz lint` to make sure and everything passed

fritz.defaults.yaml
dmitryduev: We do, just checked.

dmitryduev (approved):
LGTM! Will merge once green.

PR #167, opened 2021-02-23
Fix indentation issue in configs

kmshin1397:
Just realized that there is another issue I didn't catch in the last PR. The homepage configs are not indented enough, and they were technically not part of `skyportal.app` as they should be. Guess I fixed this at some point in `fritz.yaml` and forgot to propagate it to the defaults.


dmitryduev (approved):
Huh, nice catch!

PR #170, opened 2021-02-24
Update DBStatsHandler to use approximate photometry table size

acrellin:


stefanv (approved):

dmitryduev (approved):
👍🏻

dmitryduev:
Merging, noting (from a discussion on Slack):

> (on CI failures) it's a very annoying issue that I don't understand 100%. It seems to have to do with the vm instance that is running the workflow. Sometimes, just like here, there are docker networking issues for whatever reason, which cause the "build containers" step to fail. We ran essentially the same workflow today multiple times on the CI and it was green.

PR #186, opened 2021-03-05
Watchlists

dmitryduev:
Prompted by a request from Shri, I had a revelation on how to easily implement watchlists (see https://github.com/skyportal/skyportal/issues/1165) using our existing alert filtering infrastructure. 
This PR adds a section to the user guide with a detailed description of what needs to be done (which is essentially mostly copy-pasting).

### DIFF
--- a/doc/data/filter_examples/watchlist.json
+++ b/doc/data/filter_examples/watchlist.json
@@ -1,0 +1,264 @@
+[
+  {
+    "$project": {
+      "_id": 0,
+      "watchlist": [
+        {
+          "name": "my_favorite_spot",
+          "ra": 134.0779551,
+          "dec": -0.4421303
+        },
+        {
+          "name": "my_star",
+          "ra": 14.5551,
+          "dec": 37.4451303
+        }
+      ],
+      "max_distance_arcsec": {
+        "$literal": 2.0
+      },
+      "ra": {
+        "$degreesToRadians": "$candidate.ra"
+      },
+      "dec": {
+        "$degreesToRadians": "$candidate.dec"
+      }
+    }
+  },
+  {
+    "$project": {
+      "watchlist": 1,
+      "max_distance_arcsec": 1,
+      "dec": 1,
+      "dra_dec": {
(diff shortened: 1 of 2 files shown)


mcoughlin (commented):
This is my naivete speaking, but is there anyway to set it up such that the spherical distance calculation between two points as implemented here (obviously useful), is made to be callable as $sphericalDistance with sensible arguments? Or will folks be copying and pasting this whenver they want to try this?

acrellin (approved):
:+1:

stefanv (approved):

dmitryduev:
> This is my naivete speaking, but is there anyway to set it up such that the spherical distance calculation between two points as implemented here (obviously useful), is made to be callable as $sphericalDistance with sensible arguments? Or will folks be copying and pasting this whenver they want to try this?

@mcoughlin yea, I wish there was such a way. What I came up with here is my best option so far. 
The idea is that the user should do as little as possible to set up a watch list, so in this case it would be specifying the list, the match radius and just copy-pasting. We can potentially wrap the scary complexity of the resulting filter in a dedicated endpoint/page, such that the user only has to put in the crds/radius. (It would be somewhat similar to how we hide the complexity in K's cone- or near-searches, or how we prepend upstream aggregation stages). Wanna file an issue for that?

Speaking of the technical side of this, there is machinery on the mongo side to do computations on the sphere, but it's not directly applicable, unfortunately - that's why I just implemented the formula here.

For example, there is the [`$geoNear`](https://docs.mongodb.com/manual/reference/operator/aggregation/geoNear/) aggregation *stage*, but you can only use it for a single position. 

There's also the new [`$function`](https://docs.mongodb.com/manual/reference/operator/aggregation/function/) *operator*, which I think we'll be very useful in the future, but it also does not really hide any complexity. 

We could, in principle, implement a fake `$sphericalDistance` operator on the Fritz side with some "regex'ing", but I'd prefer to stick with only the native Mongo stuff.

mcoughlin (approved):

PR #212, opened 2021-03-29
Exit with non-zero status when tests fail

acrellin:


dmitryduev (approved):
Thank you @acrellin for the quick fix!

PR #219, opened 2021-03-31
Pin latest SkyPortal/Baselayer and Kowalski

dmitryduev:


dmitryduev:
Got a green mark on the CI so self-merging since this only pins the latest K/SP + makes a necessary change in the default config.

PR #229, opened 2021-04-11
Alerts API/frontend refactor - 1

dmitryduev:
This PR starts a series of PRs to refactor the Alerts APIs and frontend.

Concretely:

- Make the API specs more general so that it can be easily extended to survey other than ZTF in the future.
- `extensions/skyportal/skyportal/handlers/api/alert.py`:
  - Use the `penquins` library to query Kowalski. A `penquins.Kowalski` object is initialized at the module and not the handler level since in the latter case there is a lot of overhead for the unnecessary creation of such objects.
  - Merge searching by `objectId` and cone searches into a single endpoint that can handle multiple combinations of these. In the future, more filtering options can be added in addition to `objectId`.

Usage examples:

```
GET /api/alerts/ZTF21aahpzqi?instrument=ztf
GET /api/alerts?ra=171.0710824&dec=-13.8570612&radius=2&radius_units=arcsec&objectId=ZTF21aahpzqi&includeAllFields=true
GET /api/alerts?ra=171.0710824&dec=-13.8570612&radius=2&radius_units=arcsec
GET /api/alerts_aux/ZTF21aahpzqi?instrument=ztf
GET /api/alerts_cutouts/ZTF21aaposok?candid=1537220682615015003&cutout=science
GET /api/alerts_cutouts/ZTF21aaposok?candid=1537220682615015003&cutout=science&file_format=fits
```

- Frontend refactor `extensions/skyportal/static/js/components/Alerts.jsx` and `extensions/skyportal/static/js/ducks/alerts.js`:
  - Merge searching by `objectId` and cone searches into a single component. A table borrowed from @acrellin's cone-search PR is rendered with the search results:

alerts-20210411

- Update default config.

- Pin latest Kowalski - should address ingestion timeouts on the CI, see https://github.com/dmitryduev/kowalski/pull/95.


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -107,49 +118,26 @@ def get_user_streams(self):
...
+            - kowalski
           parameters:
             - in: path
               name: objectId
               required: true
               schema:
                 type: str
+            - in: query
+              name: instrument
+              required: false
```
kmshin1397, line 138:
```suggestion
              required: false
              default: 'ZTF'
```

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -107,49 +118,26 @@ def get_user_streams(self):
...
-        return resp
-
     @auth_or_token
-    async def get(self, objectId: str = None):
+    async def get(self, object_id: str = None):
         """
         ---
         single:
-          description: Retrieve a ZTF objectId from Kowalski
+          description: Retrieve an object from Kowalski by object_id
```
kmshin1397, line 126:
```suggestion
          description: Retrieve an object from Kowalski by objectId
```
I think it might be better to keep this in camel case since the parameter name will stay that way, to avoid confusion

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -226,26 +223,134 @@ async def get(self, objectId: str = None):
...
 
-        except Exception:
-            _err = traceback.format_exc()
-            return self.error(f"failure: {_err}")
+        # executing a general search
+        object_id = self.get_query_argument("objectId", None)
+        ra = self.get_query_argument("ra", None)
+        dec = self.get_query_argument("dec", None)
+        radius = self.get_query_argument("radius", None)
+        radius_units = self.get_query_argument("radius_units", None)
```
kmshin1397, line 246: The documentation for the `ZTFAlertsByCoordsHandler` got lost here I think. Should prob bring that back into the fold in the docstrings here, with these args, etc.

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -226,26 +223,134 @@ async def get(self, objectId: str = None):
...
+
+            response = kowalski.query(query=query)
+
+            if response.get("status", "error") == "success":
+                alert_data = response.get("data")
+                return self.success(data=alert_data["ZTF_alerts"]["query_coords"])
+
+            return self.error(response("message"))
+
+        # otherwise, run a find query with the specified filter
```
kmshin1397, line 326: Correct me if I'm wrong here, but we won't ever reach this block right? Either `object_id is not None` and it returns in the aggregate query case, or it goes into the positional query and either does the cone search or fails due to missing positional parameters right?

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -617,19 +722,27 @@ def post(self, objectId):
...
         ---
         single:
           description: Retrieve aux data for an objectId from Kowalski
+          tags:
+            - alerts
+            - kowalski
           parameters:
             - in: path
-              name: objectId
+              name: object_id
```
kmshin1397, line 737: Should probably decide whether we want to stick to camel or snake casing in the query params because we have both. I don't really have a preference but maybe better to stick to camel since that seems less disruptive to current users?

kmshin1397: I'm talking more about just what's actually put in the request, btw - the corresponding Python variables should definitely be snake case to match Python standards.

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -940,20 +1068,20 @@ async def get(self, objectId: str = None):
...
+            file_format = self.get_argument("file_format", "png").lower()
             interval = self.get_argument("interval", default=None)
             stretch = self.get_argument("stretch", default=None)
             cmap = self.get_argument("cmap", default=None)
 
             known_cutouts = ["Science", "Template", "Difference"]
             if cutout not in known_cutouts:
                 return self.error(
-                    f"cutout {cutout} of {objectId}/{candid} not in {str(known_cutouts)}"
+                    f"cutout {cutout} of {object_id}/{candid} not in {str(known_cutouts)}"
```
kmshin1397, line 1079:
```suggestion
                    f"Cutout {cutout} of {object_id}/{candid} not in {str(known_cutouts)}"
```

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -940,20 +1068,20 @@ async def get(self, objectId: str = None):
...
                 )
             known_file_formats = ["fits", "png"]
             if file_format not in known_file_formats:
                 return self.error(
-                    f"file format {file_format} of {objectId}/{candid}/{cutout} not in {str(known_file_formats)}"
+                    f"file format {file_format} of {object_id}/{candid}/{cutout} not in {str(known_file_formats)}"
```
kmshin1397, line 1084:
```suggestion
                    f"File format {file_format} of {object_id}/{candid}/{cutout} not in {str(known_file_formats)}"
```

kmshin1397 (commented):
Thanks @dmitryduev this refactor definitely looks like the way to proceed for the future. I still have to look through the front-end stuff, but wanted to get my comments on the backend stuff out to you.

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -68,37 +97,324 @@ const useStyles = makeStyles((theme) => ({
...
+    color: theme.palette.text.secondary,
+    position: 'absolute',
+    top: '50%',
+    left: '50%',
+    marginTop: -12,
+    marginLeft: -12,
+  },
+  grid_item_table: {
+    order: 1,
+    [theme.breakpoints.down('lg')]: {
```
kmshin1397, line 121: SkyPortal docs technically specify using only `breakpoints.up()`  [here](https://skyportal.io/docs/styling.html#responsive-design) so you should probably refactor this to have `order: 2` by default and switch to 1 at the `breakpoint.up` :)

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -68,37 +97,324 @@ const useStyles = makeStyles((theme) => ({
...
+      order: 2,
+    },
+  },
+  grid_item_search_box: {
+    order: 2,
+    [theme.breakpoints.down('md')]: {
```
kmshin1397, line 127: Same here

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -109,96 +425,66 @@ const Alerts = () => {
...
-                  >
-                    Search
-                  </Button>
+                  <div className={classes.wrapperRoot}>
+                    <div className={classes.wrapper}>
+                      <Button
+                        type="submit"
+                        variant="contained"
+                        color="primary"
+                        // className={classes.button_add}
```
kmshin1397, line 473: Delete commented out line

extensions/skyportal/static/js/components/Alerts.jsx
```diff
@@ -68,37 +97,324 @@ const useStyles = makeStyles((theme) => ({
...
+      label: "acai_b",
+      options: {
+        filter: false,
+        sort: true,
+        customBodyRender: (value, tableMeta, updateValue) => value ? value.toFixed(5) : value,
+      },
+    },
+  ];
+
+  const { register: registerForm, handleSubmit: handleSubmitForm, control: controlForm } = useForm();
```
kmshin1397, line 380: Thoughts on defining more custom validation logic on this form to check for "not all position params filled out" errors? It looks like right now, if one of those are left out, the duck will just make an `/api/alerts?objectId=${object_id}` call instead of at least passing along the ones that were filled out, meaning we won't see the more informative error messages from K right? So I think we either do that check in the form validation, or tweak the duck to something like `ra.length || dec.length || radius.length` instead of `ra.length && dec.length && radius.length`

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -226,26 +223,134 @@ async def get(self, objectId: str = None):
...
                 return self.success(data=alert_data)
             else:
-                return self.error(f"Failed to fetch data for {objectId} from Kowalski")
+                return self.error(f"Failed to fetch data for {object_id} from Kowalski")
 
-        except Exception:
-            _err = traceback.format_exc()
-            return self.error(f"failure: {_err}")
+        # executing a general search
+        object_id = self.get_query_argument("objectId", None)
```
acrellin, line 242: use different variable name than above

extensions/skyportal/skyportal/handlers/api/alert.py
acrellin: :+1:

extensions/skyportal/skyportal/handlers/api/alert.py
acrellin: These path parameters are positional, not explicitly named in the request URL, so not sure it matters/applies here @kmshin1397

kmshin1397: Ah right I missed that, you're right

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: No, it checks if `object_id` was not set in path, there are three options:
1) Only positional query was specified - this is the equivalent of @acrellin's `alerts_by_coords`
2) `objectId` was specified in query, but no positional stuff. I want to keep it this way so that in the future, we could add querying by other parameters, or multiple `objectId`'s in one go
3) Both `objectId` and positional stuff is specified. Then both are combined in the query. Again, this will be very  useful in the future when 2) is expanded.

I additionally check whether all required positional info was provided - this is where having something like `odmantic` for query spec validation that we use in K would be nice to have; but is fine as is, I guess.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: Oops, put it back.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: Fixed.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: Fixed.

dmitryduev: Fixed.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: yea, I don't think it matters here. I'd keep it camel case, not to confuse those who will be reading the API specs.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: Fixed.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: Fixed.

extensions/skyportal/static/js/components/Alerts.jsx
dmitryduev: Damn, I always forget which one it should be... I first did `up` and then thought: "oh no, it should be `down`" and rewrite it 🤣. Fixed.

extensions/skyportal/skyportal/handlers/api/alert.py
dmitryduev: It's now `object_ids` since multiple comma-separated ids are now allowed.

extensions/skyportal/static/js/components/Alerts.jsx
dmitryduev: Added a check to the `submitSearch` function. The duck shouldn't be checking such stuff and form validation would be too custom, I think.

extensions/skyportal/skyportal/handlers/api/alert.py
kmshin1397: Ah I see - I missed the fact that it was both a path and query param 👍

kmshin1397 (approved):

acrellin (approved):

PR #234, opened 2021-04-15
Try to avoid running out of database connections

stefanv:
The error:

```
QueuePool limit of size 5 overflow 10 reached, connection timed out,
timeout 30 (Background on this error at: http://sqlalche.me/e/13/3o7r)
```

This happened because we switched to a single-session-per-request
model.  Now, for each thread on the app server, there will be one or
more session (and, thus, one or more database connection).

Our current SQLAlchemy pool limits are: 5 connections,
with an overrun of 10 (i.e., a maximum of 15 database connections).
This is roughly correct: we run 8 instances of the application, and
8 * 15 = 120 (slightly larger than the 100 connections allowed to
Postgres by default).

With the new thread model, we will have at least nr_threads * nr_apps
connections.  Tornado uses asyncio, which defaults to 5 * NR_CPU
threads, and we have 8 CPUs.  Thus, 40 connections per app instance,
with 8 instances, gives 320 (way over the 100 limit).

The SQLAlchemy thread pool surfaces this error before we ever run into
problems with PostgreSQL, however.  The above is a conservative
analysis, which can potentially become worse when async is employed
inside a handler thread.

The solution proposed here is to limit the number of threads to 8 per
app, bringing us to 8 * 8 = 64 total.  When using async, each thread
can potentially have multiple handlers, but since async handlers are
not very common currently, the extra 36 connections should provide ample buffer.

### DIFF
--- a/extensions/skyportal/skyportal/app_server_fritz.py
+++ b/extensions/skyportal/skyportal/app_server_fritz.py
@@ -1,3 +1,6 @@
+import asyncio
+import concurrent
+
 from skyportal.app_server import make_app
 
 from skyportal.handlers.api.alert import (
@@ -38,6 +41,13 @@ def make_app_fritz(cfg, baselayer_handlers, baselayer_settings, process=None, en
 
     """
     app = make_app(cfg, baselayer_handlers, baselayer_settings, process, env)
+
+    # Limit the number of threads on each Tornado instance.
+    # This is to ensure that we don't run out of database connections,
+    # or overload our SQLAlchemy connection pool.
+    asyncio.get_event_loop().set_default_executor(
+        concurrent.futures.ThreadPoolExecutor(max_workers=8)
+    )
 
     app.add_handlers(r".*", fritz_handlers)  # match any host
 


extensions/skyportal/skyportal/app_server_fritz.py
```diff
@@ -39,6 +42,12 @@ def make_app_fritz(cfg, baselayer_handlers, baselayer_settings, process=None, en
     """
     app = make_app(cfg, baselayer_handlers, baselayer_settings, process, env)
 
+    # Limit each Tornado instance to 5 threads
```
acrellin, line 45:
```suggestion
    # Limit each Tornado instance to 8 threads
```

stefanv:
In discussion with Danny, another point was raised: while Tornado may be launching 8 threads (after this PR), will all 8 these threads be used, and how?  I don't have a good understanding of how Tornado implements that internally.

I am hesitant to reduce the number of threads to 1 per app: I'm worried that it may mean that we start rejecting connections once 5 have been made.  With the threading model, it may mean that we can accept around 40, which will then be served in 8 parallel chunks (one set per worker).  I suspect with our current setup (almost no async) each worker is running tasks in series.

But, that is something we can try: limit each app process to 1 thread, and then allow 1 + 5 connections in the connection pool (only 1 connection should be necessary then).

stefanv:
Update: We identified a session leak.

Each handler (request) gets its own session.  The handler is responsible
for closing the session (and, by implication, associated database connection)
when it is done.  This happens in `on_finish`.

When a handler launches a task using `add_callback`, that task could
a) access the database and b) continue operating after the handler's
`on_finish` method had been called.  This would result in the session's
database connection remaining active, and not being returned to the
SQLAlchemy thread pool.

Such connection leaks result in the error above.

One such a leak was the PS1 image loader.  We've now removed it, and we'll keep an eye out for any other situations in which this could occur.

PR #236, opened 2021-04-15
Make Fritz CLI less noisy

stefanv:
This is a start towards cleaning up the fritz CLI output.

1. Remove unnecessary prints from fire
2. Wrap dependency checks and submodule inits with `status`

This is not implemented here, but I recommend we standardize the way we call `subprocess.run` and wrap these in status calls.  This would more clearly indicate where things go wrong (currently, we just exit with an exception).

### DIFF
--- a/tools/status.py
+++ b/tools/status.py
@@ -1,15 +1,42 @@
 from contextlib import contextmanager
 import sys
+import io
 
 
 @contextmanager
-def status(message):
-    print(f"[·] {message}", end="")
+def redirect_std(out):
+    """
+    Contextmanager to temporarily redirect stdout toa StringIO object.
+
+    """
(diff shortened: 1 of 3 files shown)


tools/status.py
```diff
@@ -1,15 +1,42 @@
...
 import sys
+import io
 
 
 @contextmanager
-def status(message):
-    print(f"[·] {message}", end="")
+def redirect_std(out):
+    """
+    Contextmanager to temporarily redirect stdout toa StringIO object.
```
dmitryduev, line 9:
```suggestion
    Contextmanager to temporarily redirect stdout to a StringIO object.
```

dmitryduev (approved):
:+1:

PR #246, opened 2021-04-20
Update alert aux handler success response in docs

kmshin1397:
Current docstring is misleading - the data retrieval is successful


stefanv (approved):

dmitryduev (approved):
Oh, oops, thanks for catching that!

PR #248, opened 2021-04-24
ZTF light curves - 1

dmitryduev:
This PR starts a series of PRs that implement an interfaces to the ZTF light curve ("matchfile") data.

Concretely:
- `extensions/skyportal/skyportal/handlers/api/archive.py`:
  - `ArchiveCatalogsHandler.GET`: get list of catalogs from a Kowalski instance:
```
/api/archive_catalogs
```
  - `ArchiveHandler.GET`: get ZTF light curves (only by position for now). For example,
```
/api/archive?catalog=ZTF_sources_20210401&ra=139.7154811&dec=14.3646982&radius=2&radius_units=arcsec
```
- Frontend component with Vega plots + duck:
archive-20210424
I came up with a way to properly "colorize" the data in the plots, perhaps we might want to implement something similar for the Scanning page.
- Fix an issue in Filter.jsx, same as in https://github.com/skyportal/skyportal/pull/1909
- Fix bugs in `alert.py`

An implementation detail: before we decide on https://github.com/dmitryduev/kowalski/issues/60, since the light curve data live in a dedicated Kowalski instance ("Gloria" - there is way too much data to store in one place), we need to point SP at that other instance in the config.

TODO in the next PR (sounds simple, but it's actually _a lot_ of stuff - this PR would become way too big):
- `/api/archive.POST` to save the selected light curves to SP. Provide SP obj name
- The corresponding part on the frontend: show nearest sources already existing in SP to select from, or provide an option to create a new one. Radio group: select from existing/save new (with an auto-generated name).

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/archive.py
@@ -1,0 +1,163 @@
+from penquins import Kowalski
+
+from baselayer.log import make_log
+from baselayer.app.access import auth_or_token
+from baselayer.app.env import load_env
+from ..base import BaseHandler
+
+
+env, cfg = load_env()
+log = make_log("archive")
+
+
+gloria = Kowalski(
+    token=cfg["app.gloria.token"],
+    protocol=cfg["app.gloria.protocol"],
+    host=cfg["app.gloria.host"],
+    port=int(cfg["app.gloria.port"]),
+    timeout=10,
+)
+log(f"Gloria connection OK: {gloria.ping()}")
+
+
+class ArchiveCatalogsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        query = {"query_type": "info", "query": {"command": "catalog_names"}}
+        catalog_names = gloria.query(query=query).get("data")
+        return self.success(data=catalog_names)
+
+
+class ArchiveHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        query = {"query_type": "info", "query": {"command": "catalog_names"}}
+        # expose only the ZTF light curves for now
+        available_catalogs = [
+            catalog
+            for catalog in gloria.query(query=query).get("data")
+            if "ZTF_sources" in catalog
+        ]
+
+        # allow access to public data only by default
+        program_id_selector = {1}
+
+        for stream in self.associated_user_object.streams:
+            if "ztf" in stream.name.lower():
+                program_id_selector.update(set(stream.altdata.get("selector", [])))
+
+        program_id_selector = list(program_id_selector)
+
+        catalog = self.get_query_argument("catalog")
+        if catalog not in available_catalogs:
+            raise ValueError(f"Catalog {catalog} not available")
+
+        # executing a cone search
+        ra = self.get_query_argument("ra", None)
+        dec = self.get_query_argument("dec", None)
+        radius = self.get_query_argument("radius", None)
+        radius_units = self.get_query_argument("radius_units", None)
+
+        position_tuple = (ra, dec, radius, radius_units)
+
+        if not all(position_tuple) and any(position_tuple):
+            # incomplete positional arguments? throw errors, since
+            # either all or none should be provided
+            if ra is None:
+                return self.error("Missing required parameter: ra")
+            if dec is None:
+                return self.error("Missing required parameter: dec")
+            if radius is None:
+                return self.error("Missing required parameter: radius")
+            if radius_units is None:
+                return self.error("Missing required parameter: radius_units")
+        if all(position_tuple):
+            # complete positional arguments? run "near" query
+            if radius_units not in ["deg", "arcmin", "arcsec"]:
+                return self.error(
+                    "Invalid radius_units value. Must be one of either "
+                    "'deg', 'arcmin', or 'arcsec'."
+                )
+            try:
+                ra = float(ra)
+                dec = float(dec)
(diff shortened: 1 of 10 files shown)


dmitryduev:
@acrellin @kmshin1397 I decided to make this "ready for review". The saving logic is missing here and I will implement it in the next PR - this PR is pretty large already.

extensions/skyportal/skyportal/handlers/api/archive.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+gloria = Kowalski(
+    token=cfg["app.gloria.token"],
+    protocol=cfg["app.gloria.protocol"],
+    host=cfg["app.gloria.host"],
+    port=int(cfg["app.gloria.port"]),
+    timeout=10,
+)
+log(f"Gloria connection OK: {gloria.ping()}")
```
kmshin1397, line 20: Thoughts on naming the configuration section here something more generic? Like `kowaslki_archives` or something like that?

extensions/skyportal/skyportal/handlers/api/archive.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class ArchiveCatalogsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        query = {"query_type": "info", "query": {"command": "catalog_names"}}
```
kmshin1397, line 26: This query will need to change to filter for only the light curve catalogs if we decide to implement dmitryduev/kowalski#60, right? Maybe just leave a TODO comment here to that effect

extensions/skyportal/static/js/components/Archive.jsx
```diff
@@ -0,0 +1,541 @@
...
+import TableRow from "@material-ui/core/TableRow";
+import TableCell from "@material-ui/core/TableCell";
+import {dec_to_dms, ra_to_hours} from "../units";
+import { showNotification } from "baselayer/components/Notifications";
+
+import * as Actions from "../ducks/archive";
+import SaveIcon from "@material-ui/icons/Save";
+import {IconButton} from "@material-ui/core";
+import HelpOutlineIcon from "@material-ui/icons/HelpOutline";
+import Popover from "@material-ui/core/Popover";
```
kmshin1397, line 35: Sorry, this is a little picky but would be nice if the imports were more organized here, i.e. all the MUI ones together and then all the other SP imports grouped together

extensions/skyportal/static/js/components/Archive.jsx
```diff
@@ -0,0 +1,541 @@
...
+    // eslint-disable-next-line no-underscore-dangle
+    const ZTFLightCurveId = ztf_light_curves[rowMeta.dataIndex]._id;
+    const ZTFLightCurveFilterId = ztf_light_curves[rowMeta.dataIndex].filter;
+    const ZTFLightCurveData = ztf_light_curves[rowMeta.dataIndex].data.map(
+      obj => ({ ...obj, filter: ZTFLightCurveFilterId })
+    );
+    const colorScale = {
+      domain: [ZTFLightCurveFilterId],
+      range: [ZTFLightCurveColors[ZTFLightCurveFilterId]],
+    }
```
kmshin1397, line 236: Nice!

extensions/skyportal/static/js/components/VegaPlotZTFArchive.jsx
```diff
@@ -0,0 +1,191 @@
...
+        opacity: {
+          condition: { selection: "filterErrBars", value: 1 },
+          value: 0,
+        },
+      },
+    },
+  ],
+});
+
+const VegaPlot = withOrientationChange(({ data, color_scale, isPortrait }) => (
```
kmshin1397, line 156: Any reason `color_scale` isn't in camel case?

kmshin1397 (commented):
Looking good here, @dmitryduev! A couple of minor comments:

extensions/skyportal/static/js/components/Archive.jsx
dmitryduev: Yea, I've been treating js code as simply very long spells that eventually do magic once you've added the proper ingredient to the cauldron... Fixed.

extensions/skyportal/static/js/components/VegaPlotZTFArchive.jsx
dmitryduev: No reason, fixed.

extensions/skyportal/skyportal/handlers/api/archive.py
dmitryduev: I think the best will be to somehow implement https://github.com/dmitryduev/kowalski/issues/60 so that a separate entry in the config is unnecessary. I'd leave it as is for now and add comments with explanations to the code.
A side note that it's a pretty internal thing at this point: matchfiles are not public; light curves _are_ made public, but we don't have a reader/ingester for that (just filed https://github.com/dmitryduev/kowalski/issues/107).

extensions/skyportal/skyportal/handlers/api/archive.py
dmitryduev: No, this should return everything available, in case we decide to expose more of K's functionality through Fritz. The specific handlers below (like `ArchiveHandler.GET`) filter out the unnecessary stuff.

acrellin (approved):

kmshin1397 (approved):

PR #250, opened 2021-04-29
Update DB queries

acrellin:
@dmitryduev I did part of the alert handler here -- just showing you what it'd look like so far. Shall I leave the rest for you to do, as you requested?


acrellin:
Note DB stats handler requires "System admin" ACL so the queries there can remain as-is, as a system admin can view any and all records.

dmitryduev (approved):
@acrellin thanks! I'll try to get to rewriting the AlertHandler.POST guy later today once 248 and 249 are in - will ping you then.

PR #268, opened 2021-05-15
Update TNS info fetching logic & redux state structure

acrellin:
The updated state structure allows us to easily track whether the state corresponds to the source being viewed or whether it needs to be re-fetched.


dmitryduev (approved):
Good thought!

PR #269, opened 2021-05-15
Fix TNS name display

dmitryduev:
+ take into account that there could potentially be multiple matches, although that is rare.

Proof that it actually works:  :)
tns-1

### DIFF
--- a/extensions/skyportal/static/js/components/TNSInfo.jsx
+++ b/extensions/skyportal/static/js/components/TNSInfo.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect } from "react";
 import { useSelector, useDispatch } from "react-redux";
 import PropTypes from "prop-types";
 
@@ -9,7 +9,7 @@ const TNSInfo = ({ objID }) => {
   const tnsInfo = useSelector((state) => state.tnsInfo);
 
   useEffect(
-    function fetchTNSInfo() {
+    () => {
       if (tnsInfo === null || !Object.keys(tnsInfo).includes(objID)) {
         dispatch(tnsInfoActions.fetchTNSInfo(objID));
       }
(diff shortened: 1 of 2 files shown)


extensions/skyportal/static/js/components/TNSInfo.jsx
```diff
@@ -9,7 +9,7 @@ const TNSInfo = ({ objID }) => {
   const tnsInfo = useSelector((state) => state.tnsInfo);
 
   useEffect(
-    function fetchTNSInfo() {
```
acrellin, line 12: We want these to be named functions actually, please revert

extensions/skyportal/static/js/components/TNSInfo.jsx
```diff
@@ -21,23 +21,23 @@ const TNSInfo = ({ objID }) => {
...
+        objTnsInfo.map((TNSMatch) => (
           <a
-            href={`https://www.wis-tns.org/object/${objTnsInfo.name.split(" ")[1]}`}
+            key={TNSMatch.name}
+            href={`https://www.wis-tns.org/object/${TNSMatch.name.split(" ")[1]}`}
             target="_blank"
             rel="noopener noreferrer"
           >
-            {objTnsInfo.name}
+            {`${TNSMatch.name} `}
```
acrellin, line 36:
```suggestion
            TNSMatch.name
```

acrellin (commented):
Still could be good to ensure that it's a string before calling split on it so it doesn't crash

extensions/skyportal/static/js/components/TNSInfo.jsx
dmitryduev: I wanted a whitespace after the name in case there were multiple matches, which is very rare, but could happen in principle. Do you have a better suggestion for that?

extensions/skyportal/static/js/components/TNSInfo.jsx
dmitryduev: Done. [It was pycharm's suggestion.]

dmitryduev: Ah, actually ESLint complained, so I disabled it.

acrellin (approved):
Nice, thanks!

PR #298, opened 2021-06-08
Clean up group_ids logic in alerts handler

acrellin:
Previously, group_ids would only ever default to the user's groups if there was an error casting the elements to ints.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -620,7 +620,7 @@ class AlertHandler(BaseHandler):
                     ]
                 except ValueError:
                     return self.error(
-                        "Invalid group_ids parameter: all elements must " "be integers."
+                        "Invalid group_ids parameter: all elements must be integers."
                     )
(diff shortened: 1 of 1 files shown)


kmshin1397 (commented):
I can see why this is better than what is there now, but do you know why the specific issue in the beta testing channel arose, @acrellin? It seems to me that given that they say they passed in `params = {"candid": 1,"group_ids": [254]}`, it would have needed to thrown an exception in that `try...except` block for that to have resulted in being saved to all groups and I'm not sure what the exception would have been...

acrellin:
@kmshin1397 Yeah I just figured out that the issue there was that they were passing the payload to `params` instead of `json`, and those are JSON body params, not query/URL params.

kmshin1397 (approved):

kmshin1397:
> @kmshin1397 Yeah I just figured out that the issue there was that they were passing the payload to `params` instead of `json`, and those are JSON body params, not query/URL params.

Ah I see, that makes sense. Then I'm happy to approve this 👍

PR #300, opened 2021-06-09
Alert handler updates: make group_ids required & clean up group_ids checking logic

acrellin:


kmshin1397 (commented):
I agree with making `group_ids` required, but thoughts on perhaps allowing a `group_ids=all` option instead of requiring it to be a specific list of group IDs? Maybe having an `all` options wouldn't be missed, but I do have some hesitation of removing easy access to that option entirely. Perhaps @dmitryduev has some more insight into whether that's something the users would need or not.

acrellin:
Yeah, that sounds like good -- will update accordingly

acrellin:
@kmshin1397 I've updated this to allow `group_ids="all"`

kmshin1397 (approved):

PR #306, opened 2021-06-25
Sources from external catalogs in Offset plot

dmitryduev:
In this PR:

- An API endpoint to run a cross-match query against all catalogs available on Kowalski/Gloria (i.e. same instance that has the ZTF light curves), e.g. 

```/api/archive/cross_match?ra=139.7154811&dec=14.3646982&radius=2&radius_units=arcsec```

- Display *all* cross-matches within 10" from the source position on the Centroid Plot on the Source page. Tweaked a few things such as the plot size, the color scheme (so that in particular the ZTF data are properly mapped to colors), and the plot axes aspect ratio (now will be guaranteed to be 1:1, wasn't the case previously). Using different shapes for ZTF points and x-match on the same Vega lite plot proved extremely painful (and I'm not even sure it's possible to e.g. show all ZTF data as circles and all x-matches as stars), so I just made them of different size.

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

UPDATE:
- Will also display distances to closest sources in arcsec, per catalog (see a screenshot below).

### DIFF
--- a/extensions/skyportal/skyportal/app_server_fritz.py
+++ b/extensions/skyportal/skyportal/app_server_fritz.py
@@ -24,9 +24,9 @@ fritz_handlers = [
     (r"/api/alerts_aux(/.+)?", AlertAuxHandler),
     (r"/api/alerts_cutouts(/.+)?", AlertCutoutHandler),
     # Archive
-    (r"/api/archive_catalogs", ArchiveCatalogsHandler),
     (r"/api/archive", ArchiveHandler),
-    (r"/api/cross_match", CrossMatchHandler),
+    (r"/api/archive/catalogs", ArchiveCatalogsHandler),
+    (r"/api/archive/cross_match", CrossMatchHandler),
     # Alert Stream filter versioning via K:
     (r"/api/filters/([0-9]+)?/v", KowalskiFilterHandler),
     (r"/api/tns_info/(.*)", TNSInfoHandler),
--- a/extensions/skyportal/skyportal/app_server_fritz.py
+++ b/extensions/skyportal/skyportal/app_server_fritz.py
@@ -11,6 +11,7 @@ from skyportal.handlers.api.alert import (
 from skyportal.handlers.api.archive import (
     ArchiveCatalogsHandler,
     ArchiveHandler,
+    CrossMatchHandler,
 )
 from skyportal.handlers.api.kowalski_filter import KowalskiFilterHandler
 from skyportal.handlers.api.tns_info import TNSInfoHandler
@@ -25,6 +26,7 @@ fritz_handlers = [
     # Archive
     (r"/api/archive_catalogs", ArchiveCatalogsHandler),
     (r"/api/archive", ArchiveHandler),
+    (r"/api/cross_match", CrossMatchHandler),
     # Alert Stream filter versioning via K:
     (r"/api/filters/([0-9]+)?/v", KowalskiFilterHandler),
     (r"/api/tns_info/(.*)", TNSInfoHandler),
--- a/extensions/skyportal/skyportal/handlers/api/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/archive.py
@@ -361,9 +361,10 @@ class CrossMatchHandler(BaseHandler):
                     catalog: query_coords["query_coords"]
                     for catalog, query_coords in response.get("data").items()
                 }
-                # normalize positional data
+                # stringify _id's and normalize positional data
                 for catalog, sources in data.items():
                     for source in sources:
+                        source["_id"] = str(source["_id"])
                         source["ra"] = (
                             source["coordinates"]["radec_geojson"]["coordinates"][0]
                             + 180
--- a/extensions/skyportal/skyportal/handlers/api/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/archive.py
@@ -276,11 +276,11 @@ class CrossMatchHandler(BaseHandler):
                 schema: Error
         """
         query = {"query_type": "info", "query": {"command": "catalog_names"}}
-        # expose all but the ZTF-related catalogs
+        # expose all but the ZTF/PTF-related catalogs
         catalogs = [
             catalog
             for catalog in gloria.query(query=query).get("data")
-            if not catalog.startswith("ZTF")
+            if not catalog.startswith("ZTF") and not catalog.startswith("PTF")
         ]
         if len(catalogs) == 0:
             return self.error("No catalogs available to run cross-match against.")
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

File: extensions/skyportal/skyportal/tests/api/test_archive.py [hunk 1]
```diff
@@ -12,6 +12,24 @@ def test_get_archive_catalog(view_only_token):
...
+    ra, dec = 0.00017675657877, 80.01266744553764
+    radius, radius_units = 2, "arcsec"
+
+    status, data = api(
+        "GET",
+        f"cross_match?ra={ra}&dec={dec}"
+        f"&radius={radius}&radius_units={radius_units}",
+        token=view_only_token,
+    )
+    print(status, data)
```
kmshin1397, line 26:
```suggestion
```

extensions/skyportal/skyportal/app_server_fritz.py
```diff
@@ -25,6 +26,7 @@
     # Archive
     (r"/api/archive_catalogs", ArchiveCatalogsHandler),
     (r"/api/archive", ArchiveHandler),
+    (r"/api/cross_match", CrossMatchHandler),
```
kmshin1397, line 29: Thoughts on renaming the route to `/api/archive/cross_match`? Mostly since the `CrossMatchHandler` is under the `archive.py` umbrella. Seems to make a tiny bit more sense but I don't really have a preference one way or the other. Would make sense to go with `/api/archive/catalogs` as well in that case.

extensions/skyportal/static/js/components/CentroidPlot.jsx
```diff
@@ -0,0 +1,495 @@
...
+  return { delRA, delDec };
+};
+
+// Temporary function for draft
+const getReferencePoint = (ras, decs) => {
+  const refRA = d3.median(ras);
+  const refDec = d3.median(decs);
+
+  return { refRA, refDec };
+};
```
kmshin1397, line 54: I vaguely remember that this wasn't originally what the GROWTH marshal was doing in terms of reference point for the offsets, to just take the median? I think this was what we implemented since we didn't have the crossmatch info before. Do we still want the reference to just be the median of the data instead of the nearest crossmatch or whatever it was for the GROWTH marshal?

kmshin1397 (commented):
A few minor questions but looks good overall!

extensions/skyportal/skyportal/app_server_fritz.py
dmitryduev: Great idea, made the change.

extensions/skyportal/skyportal/tests/api/test_archive.py
dmitryduev: Fixed.

extensions/skyportal/static/js/components/CentroidPlot.jsx
dmitryduev: To my taste, this makes the most sense: the reference point will be calculated using all ZTF photometry (which is usually a lot of points) and then we'll show all the cross-matched sources from other catalogs within 10" (could be many) relative to that.

kmshin1397 (approved):

extensions/skyportal/static/js/components/CentroidPlot.jsx
kmshin1397: Sure makes sense to me

extensions/skyportal/skyportal/handlers/api/archive.py
```diff
@@ -223,6 +223,149 @@ def get(self):
...
+            name: radius
+            required: true
+            schema:
+              type: float
+            description: Max distance from specified (RA, Dec) (capped at 1 deg)
+          - in: query
+            name: radius_units
+            required: true
+            schema:
+              type: string
```
profjsb, line 258:
```suggestion
              type: string
              enum: [deg, arcmin, arcsec]
```

extensions/skyportal/skyportal/handlers/api/archive.py
```diff
@@ -223,6 +223,149 @@ def get(self):
...
+            required: true
+            schema:
+              type: float
+            description: Dec. in degrees
+          - in: query
+            name: radius
+            required: true
+            schema:
+              type: float
+            description: Max distance from specified (RA, Dec) (capped at 1 deg)
```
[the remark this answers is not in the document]
dmitryduev: Fixed.

extensions/skyportal/skyportal/handlers/api/archive.py
dmitryduev: Fixed.

dmitryduev:
@profjsb I added storing available K catalog names at the module level - the chances of someone adding a new catalog during its lifetime are really low.

dmitryduev:
@profjsb on a second thought, I removed the dumb module-level caching since it screwed a test: the app spins up before any data are ingested into K on the CI, so something more clever would be required. Actually that query is so cheap that I wouldn't worry about it much.

extensions/skyportal/static/js/components/CentroidPlot.jsx
profjsb: There's an implicit assumption here that the uncertainties are the same for all the photometry points. Should consider a weighted average (after sigma clipping).

Also, how will this snippet behave if there are any nans, none, etc. in `ras`, `decs`

profjsb (changes requested):
See comment about handling nan, none in the median (https://github.com/fritz-marshal/fritz/pull/306#discussion_r669000988). I had to handle this in the periodogram work on the front end (cf. https://github.com/skyportal/skyportal/blob/32939eb7f496479994f4ad21e19f920965fddf51/static/js/components/Periodogram.jsx#L106). 

If it's not too painful it would be useful to display the distance to the nearest source from each catalog in arcsec. E.g. Gaia (0.11+-0.02"), PS1 (1.20+-0.15"), ...

extensions/skyportal/static/js/components/CentroidPlot.jsx
dmitryduev: @profjsb Regarding nans/nones: we preselect only the photometry data points with non-null RA/Decl, see https://github.com/fritz-marshal/fritz/pull/306/files#diff-3608c51f5312a907615fed7fcbc9f46b0cff6700b05032f7f51a58d21aa6517cR341. 

As for the uncertainties, ZTF alerts contain SExtractor'ed parameters of a PSF-fit, such as the FWHM (see https://github.com/ZwickyTransientFacility/ztf-avro-alert/blob/master/schema/candidate.avsc#L37), which are not converted to uncertainties in RA/Dec's. We don't do that on the K side, and only post RA/Dec's (see https://github.com/dmitryduev/kowalski/blob/master/kowalski/alert_broker_ztf.py#L1165). We probably should, but I think it's outside of the scope of this PR.

dmitryduev:
@profjsb Done:
image

profjsb (approved):

PR #313, opened 2021-07-07
Fix bug when searching alerts from source page

dmitryduev:
In this PR: 
- Addressed a bug where clicking on the Search ZTF Alert Archive on the Source Page didn't work.
- Rearranged the buttons on the Source page a bit so that they take less space.
search-alerts
- Fixed a bug in test_get_alert_cutout.

### DIFF
--- a/.github/workflows/build-test-doc_upload.yaml
+++ b/.github/workflows/build-test-doc_upload.yaml
@@ -32,8 +32,8 @@ jobs:
     - name: Build docs
       run: |
         pip install wheel
-        pip install -r skyportal/baselayer/requirements.txt
         pip install -r skyportal/requirements.txt
+        pip install -r skyportal/baselayer/requirements.txt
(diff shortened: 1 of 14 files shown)


acrellin:
What's the bug fix here? Removing the prop from `propTypes`?

dmitryduev:
@acrellin my duck stupidly assumed that ra/dec/radius would only come from the alert page in the form of strings. I just fixed that assumption.

dmitryduev:
This: https://github.com/fritz-marshal/fritz/pull/313/files#diff-1fdaef78be891e5f488d9148f5579148867874179a8fdcd117bd6e145e3b9a2cR11

dmitryduev:
Hmm, failure on the CI looks real, never seen it before, and it happened in an unexpected place. Investigating.

acrellin (approved):

PR #314, opened 2021-07-13
Update SP

acrellin:


dmitryduev:
@acrellin there are a couple other things that need to pinned/fixed, I think. I may have addressed those in https://github.com/fritz-marshal/fritz/pull/306, let's see what the CI has to say about it :)

PR #315, opened 2021-07-15
Update K and SP - 20210715

dmitryduev:


acrellin:
@dmitryduev I propose we close this in favor of #316 which has a newer SP and expanded config

acrellin:
(I didn't see this one til after I opened the newer one ;) )

PR #316, opened 2021-07-16
Update SP & add rate-limiting params to config

acrellin:


dmitryduev (commented):
Thanks a lot @acrellin! Looks like SP was still rate-limiting K's requests, see:
https://github.com/fritz-marshal/fritz/pull/316/checks?check_run_id=3107797626#step:7:2177

We may have to get the gateway IP address manually and propagate it into the config, see
https://github.com/fritz-marshal/fritz/blob/dd5f5bb1c77003f90a9c262b7d049f08027f1093/fritz#L116

Did the test pass locally? I'd make sure K's IP is within the whitelisted range...

dmitryduev:
Ah, the fix could be as simple as pinning the latest K - could you please do that @acrellin?

stefanv:
Should be safe to whitelist the following by default:

```
10.0.0.0/8
172.16.0.0/12
192.168.0.0/16
```

acrellin:
@dmitryduev this already pins K d8c5a86

dmitryduev (approved):

PR #322, opened 2021-07-28
Pin latest SP & K; update fritz.defaults.yaml rate-limiting IP addresses

acrellin:


dmitryduev (approved):
Good to (self-)merge!

PR #323, opened 2021-07-29
Add basic ACAI plot to source page + Update CentroidPlot.jsx

### LINKED ISSUE #247: Display ACAI values graphically
**Feature Summary**
Visual display of ACAI values like Alerce's stamp/LC classifiers.
Will make interpretation quicker. 
Suggested titles for the ACAI pentagon:
hosted, nuclear, orphan, "real", variable

**Usage / behavior**
A static figure based on latest ACAI values (or median ACAI values)

**Additional context**
[For an example, see this pentagonal figure](https://alerce.online/object/ZTF18aaaeyfq)
alerce_stamp_classifier_display

kmshin1397:
Adds a (very) basic ACAI values plot to the Source page, simply plotting the latest value found in the loaded source's annotations for each program in a bar chart:
acai

Also brings in the CentroidPlot.jsx and HR diagram styling changes introduced in https://github.com/skyportal/skyportal/pull/2154

Of course, much more could be done with this as future improvements, i.e. querying K for the full series of ACAI values and producing more detailed plots with medians, std dev, etc. This serves more as a basic starting point to get something at least somewhat useful out there. Seems like this should at least address the extent of the functionality in the original feature request.

Closes #247

### DIFF
--- a/extensions/skyportal/static/js/components/AcaiPlot.jsx
+++ b/extensions/skyportal/static/js/components/AcaiPlot.jsx
@@ -1,0 +1,187 @@
+import React, { useState } from "react";
+import { useSelector } from "react-redux";
+import PropTypes from "prop-types";
+import Tabs from "@material-ui/core/Tabs";
+import Tab from "@material-ui/core/Tab";
+import Box from "@material-ui/core/Box";
+import { makeStyles, useTheme } from "@material-ui/core/styles";
+import embed from "vega-embed";
+
+const useStyles = makeStyles(() => ({
+  container: { width: "100%" },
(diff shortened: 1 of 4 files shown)


dmitryduev:
Closing bc:
- We have moved the Centroid plot stuff into a different PR that is now merged.
- The ACAI plot probably requires some re-engineering: users don't care that much about the individual latest values per filter (which could all be the same if executed on the same alert). In a future PR, would be useful to query K and show acai scores as a function of time for a given objectId + a SCoPe-style score distribution plot (see https://arxiv.org/pdf/2102.11304.pdf Fig. 10).

PR #348, opened 2021-09-20
Enable cronjob to delete old, unsaved candidates

acrellin:


fritz.defaults.yaml
```diff
@@ -352,9 +352,8 @@ skyportal:
   cron:
     - interval: 60
       script: jobs/count_unsaved_candidates.py
-    # - interval: 1440
-    #   script: jobs/delete_unsaved_candidates.py
-    #   limit: ["01:00", "02:00"]
```
acrellin, line 357: Unless it matters what time we run this?

dmitryduev:
@acrellin thanks for this. You know what, I think we should first do this manually -- there are tons and tons of unsaved candidates, so cleaning up photometry will potentially take a long time, so we need to be careful about that. I suggest to discuss the details at the iteration meeting this week.

mcoughlin:
Given the age of this PR and that things seem to run OK as is, I will close this for now. Feel free to reopen as needed.

PR #350, opened 2021-09-24
Update Centroid plot

dmitryduev:
This is a port of @kmshin1397's #323 (see for the details) that only updates the Centroid plot (brings in the CentroidPlot.jsx and HR diagram styling changes introduced in skyportal/skyportal#2154), since we still haven't decided what we wanted for the ACAI classifications from that PR.

### DIFF
--- a/extensions/skyportal/static/js/components/CentroidPlot.jsx
+++ b/extensions/skyportal/static/js/components/CentroidPlot.jsx
@@ -105,7 +105,7 @@ const getMessages = (delRaGroup, delDecGroup) => {
 };
 
 // The Vega-Lite specifications for the centroid plot
-const spec = (inputData) => ({
+const spec = (inputData, textColor) => ({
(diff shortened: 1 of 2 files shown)


dmitryduev:
BTW @acrellin, looking at the diffs -- perhaps it would be a good idea to bring in optional chaining for filters and maps throughout the SP extensions in F, like you did in SP.

acrellin:
@dmitryduev these all look pretty safe to me, actually; all of the `.map` and `.filter` uses appear to start with `Object.keys(..)`, which will either always produce an array or will error if provided an invalid argument.

acrellin (approved):

dmitryduev:
@acrellin I meant all filters and maps, not just what's here - again some potentially formidable grepwork ;)
mind merging? soo green...

acrellin:
@dmitryduev ah I see, yeah I've been thinking about doing that for the whole extensions codebase... Will file an issue and try to do it soon!

Issue #399, opened 2022-03-11
A readme file with example use of Fritz-API calls 


janvanroestel:
**Feature Summary**
A simple readme that collects links to github pages of users. The goal is to avoid duplication of work and a starting point for new development.

mcoughlin:
Closing as stale: several years old and no longer applicable to the current Fritz/BOOM architecture. Please refile if still relevant.

PR #400, opened 2022-03-11
added a doc user_APIexamples.md

janvanroestel:
I added a doc where we can collect links to user-examples. As mentioned, this is to avoid duplication of work; emailing all Fritz users to ask if someone has done something already etc. 

This should be linked to issues #399

### DIFF
--- a/doc/user_APIexamples.md
+++ b/doc/user_APIexamples.md
@@ -1,0 +1,14 @@
+# Fritz API examples
+
+[Fritz](https://github.com/fritz-marshal/fritz) has a fully functional API. The documentation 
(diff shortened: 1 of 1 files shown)


guynir42:
This seems like an old PR but probably worth putting it in. 
My main concern is that maybe we should put it into another docs file? 
Any thoughts on this @profjsb ?

guynir42:
@profjsb Do you think this can be merged? Should we maybe change the filename?

mcoughlin:
Given the age of this PR and that things seem to run OK as is, I will close this for now. Feel free to reopen as needed.

PR #402, opened 2022-03-17
prepare config for sncosmo upgrade

mcoughlin:
This PR removes the additional_bandpasses bandpasses column given those (correct) bandpasses appear in sncosmo 2.8. To be merged jointly with:
https://github.com/skyportal/skyportal/pull/2763


mcoughlin:
Based on the discussion today, I will merge this now.

PR #410, opened 2022-04-29
Remove ResizeObserver from source page

jadalilleboe:
Same changes as in [#2888](https://github.com/skyportal/skyportal/pull/2888)


jadalilleboe:
I would think the failure is unrelated here?

PR #425, opened 2022-06-19
add galaxy to fritz sidebar

mcoughlin:
This PR adds the galaxy page to the fritz sidebar.


guynir42 (approved):
LGTM. I am wondering if it may be better to put together all the GCN related pages into one folder on the sidebar.

mcoughlin:
@guynir42 We could think about it. Right now the thinking is that the dropdown is stuff you don't interact with much, so like uploads and such. But maybe to be discussed on a future call.

PR #433, opened 2022-06-27
Update Source Page as in Skyportal

thomasculino:
Because of some extra buttons, SourceDesktop.jsx and SourceMobile.jsx were not updated with skyportal's new modifications. This PR aims to make these modifications manually as requested in https://github.com/skyportal/skyportal/pull/3143.


mcoughlin (approved):
@thomasculino Thanks!

Issue #438, opened 2022-07-01
alerts_aux Session error


mcoughlin:
I think https://github.com/fritz-marshal/fritz/blob/main/extensions/skyportal/skyportal/handlers/api/alert.py#L224 maybe just needs a Session?

**Describe the bug**
  File "/skyportal_env/lib/python3.8/site-packages/tornado/web.py", line 1702, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/skyportal/baselayer/app/access.py", line 46, in wrapper
    return method(self, *args, **kwargs)
  File "/skyportal/skyportal/handlers/api/alert.py", line 858, in get
    for stream in self.associated_user_object.streams:
  File "/skyportal_env/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py", line 481, in __get__
    return self.impl.get(state, dict_)
  File "/skyportal_env/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py", line 941, in get
    value = self._fire_loader_callables(state, key, passive)
  File "/skyportal_env/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py", line 977, in _fire_loader_callables
    return self.callable_(state, passive)
  File "/skyportal_env/lib/python3.8/site-packages/sqlalchemy/orm/strategies.py", line 861, in _load_for_state
    raise orm_exc.DetachedInstanceError(
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <User at 0x7fc4e24fb220> is not bound to a Session; lazy load operation of attribute 'streams' cannot proceed (Background on this error at: https://sqlalche.me/e/14/bhk3)

**To Reproduce**
 https://fritz.science/api/alerts_aux/ZTF22aaqbbkz?includePrvCandidates=False1.

PR #439, opened 2022-07-01
Fix Alert user detached from session

guynir42:
There's a bug when using the alerts handler, when the current user cannot access streams because it is detached form a session. 
I opened a `self.Session` that should attach the current user to the new session and let it load streams.


guynir42:
Addresses #438

mcoughlin (approved):
@guynir42 Thanks for getting this one!

PR #441, opened 2022-07-07
La Palma longitudes should be west

ebellm:


mcoughlin:
@ebellm Could you take a quick look at the SkyPortal equivalent to see if anything is missing / wrong? https://github.com/skyportal/skyportal/blob/main/data/telescopes.yaml that will be the default loaded going forward.

ebellm:
@mcoughlin The SkyPortal version looks okay--I note that the SkyPortal version intermixes -180/180 and 0-360 longitudes.

mcoughlin:
@ebellm yeah a bit annoying, but at least it's working for now. Thanks for checking.

PR #442, opened 2022-07-07
Update fritz config with analysis paths

mcoughlin:
This PR adds the analysis paths to the fritz config.

### DIFF
--- a/fritz.defaults.yaml
+++ b/fritz.defaults.yaml
@@ -117,6 +117,8 @@ skyportal:
         component: TelescopePage
       - path: "/instruments"
         component: InstrumentPage
+      - path: "/services"
+        component: AnalysisServicePage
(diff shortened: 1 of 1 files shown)


guynir42 (approved):
Were there no changes to the source page as well?

mcoughlin:
@guynir42 that will come once the SP PR is merged.

guynir42:
@mcoughlin does it make sense to merge this before the SP PRs are in?

guynir42:
What I'm wondering is would it be useful to the users to have an analysis page if they don't have all the tools to run it on objects. Is the analysis page usable on its own?

mcoughlin:
@guynir42 Yes. This deals with the concept of an analysis service. The other PR adds a front end to trigger them (but API still is fine).

PR #445, opened 2022-07-11
Fix User not bound to session in archive handler

guynir42:


mcoughlin (approved):
Thanks @guynir42. Will be good to try this out.

PR #453, opened 2022-08-05
Add missing skyportal configs to fritz

mcoughlin:
This PR adds missing skyportal configs to fritz.

### DIFF
--- a/fritz.defaults.yaml
+++ b/fritz.defaults.yaml
@@ -126,6 +126,24 @@ skyportal:
       - path: "/followup_requests"
         component: FollowupRequestPage
 
+    gcn:
+      server: gcn.nasa.gov
+      # you can obtain a client_id and client_secret at https://gcn.nasa.gov/quickstart
+      client_id:
+      client_secret:
+      notice_types:
+        - FERMI_GBM_FLT_POS
+        - FERMI_GBM_GND_POS
+        - FERMI_GBM_FIN_POS
+        - FERMI_GBM_SUBTHRESH
+        # - LVC_PRELIMINARY
+        # - LVC_INITIAL
+        # - LVC_UPDATE
+        # - LVC_RETRACTION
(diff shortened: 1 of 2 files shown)


fritz.defaults.yaml
```diff
@@ -51,6 +51,8 @@ skyportal:
 
     treasuremap_endpoint: http://treasuremap.space
 
+    tns_endpoint: https://sandbox.wis-tns.org
```
guynir42, line 54: Is it possible that this was deliberately removed from the fritz config (used in staging) to prevent accidentally posting things to TNS?

fritz.defaults.yaml
```diff
@@ -436,6 +459,16 @@ skyportal:
     cosmology: Planck18
     #cosmology: {H0: "65.0", Om0: 0.3, Ode0: 0.7, name: 'skyportal_user_cosmo'}
 
+    # The minimum signal-to-noise ratio/ n-sigma for lim mag cacluations to
+    # consider a photometry point as a detection
+    photometry_detection_threshold_nsigma: 3.0
```
guynir42, line 464: Doesn't this already exist in the skyportal config?

mcoughlin: Yes but since its a subsection it doesn't merge.

fritz.defaults.yaml
guynir42: I guess. Is that going to help make some error go away?

mcoughlin: yeah it was making it impossible for docs to build and such

fritz.defaults.yaml
guynir42: I don't mind adding it but it is likely going to get deleted when we finally get the gang together and figure out the whole merging config thing.

guynir42 (approved):

PR #455, opened 2022-08-09
Add nmma as a fritz extension

mcoughlin:
This PR adds nmma as a fritz extension.


Theodlz:
Closed in favor of NMMA API running as an analysis service in the cloud

PR #456, opened 2022-08-12
Add analysis pages

mcoughlin:
This PR adds analysis pages to the fritz config.


guynir42 (approved):
Are you sure there weren't any other definitions that needed to go into the config? Or were they successfully merged from the skyportal config?

mcoughlin:
@guynir42 I think it's just the path. The other things are demo_data.

PR #466, opened 2022-09-12
Refactor alert handler

mcoughlin:
This PR refactors the alert handler to simplify calls from other handlers.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -109,9 +109,9 @@ def make_thumbnail(a, ttype, ztftype):
     return thumb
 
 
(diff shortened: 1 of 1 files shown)


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -108,6 +109,344 @@ def make_thumbnail(a, ttype, ztftype):
     return thumb
 
 
+def post_alert(objectId, candid, group_ids, program_id_selector, user_id, session):
```
guynir42, line 112:
```suggestion
def post_alert(object_id, candid, group_ids, program_id_selector, user_id, session):
```

guynir42 (approved):
LGTM

PR #467, opened 2022-09-13
Add optional plugins for the source and candidate page

mcoughlin:
This PR adds the "plugins" for the source and candidate page.


guynir42 (approved):
Great!

PR #470, opened 2022-09-23
Add alert thumbnail retrieval

mcoughlin:
This PR adds a button to specifically retrieve thumbnails for an object.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -1347,17 +1347,16 @@ class AlertTripletsHandler(BaseHandler):
             default: false
         responses:
           '200':
-            description: retrieved cutout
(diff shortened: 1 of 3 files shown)


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -437,6 +452,7 @@ def post_alert(object_id, candid, group_ids, program_id_selector, user_id, sessi
             cutout = dict()
 
         thumb = make_thumbnail(cutout, ttype, ztftype)
+        post_thumbnail(thumb, user_id, session)
 
         try:
             post_thumbnail(thumb, user_id, session)
```
guynir42, line 458: I don't think you want to do this twice...

guynir42 (approved):
Minor comment. This will be really useful, thanks!

PR #472, opened 2022-10-01
Change Archive -> Persistent Sources in panel

mcoughlin:
This PR changes Archive to Persistent Sources in the left panel, which is a much more intuitive name.


guynir42 (approved):
Better, I guess.

PR #475, opened 2022-10-07
Remove slow db_stats queries

mcoughlin:
For a large DB, the candidates timing queries are prohibitive (and I don't see the use to them). So let's remove them.


guynir42 (approved):
LGTM

PR #479, opened 2022-10-12
add alert_available

guynir42:
This combines with skyportal PR #3534 to allow importing alert.py without a try/catch statement.


profjsb (approved):
LGTM

PR #481, opened 2022-10-14
Fix case where alert dataframe has no candid

guynir42:
When posting an alert from the `ThumbnailsPathHandler` I get errors where there is no "candid" key in the dataframe returned from `alert_data`. 

I am not sure why this is happening but I can't regenerate the thumbnails this way.


mcoughlin (approved):
LGTM

guynir42:
This needs a more in depth dive into fixing `post_alert`

PR #482, opened 2022-10-15
Default arguments for post_alert

mcoughlin:
This PR adds default arguments for post_alert.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -262,7 +262,6 @@ def post_alert(
             alert_data["prv_candidates"].append(latest_alert_data["candidate"])
 
     df = pd.DataFrame.from_records(alert_data["prv_candidates"])
-    candid = None
     if candid is None:
         candid = int(latest_alert_data["candidate"]["candid"])
 
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -265,6 +265,8 @@ def post_alert(
             alert_data["prv_candidates"].append(latest_alert_data["candidate"])
 
     if candid is None:
+        if len(latest_alert_data) == 0:
+            raise ValueError("Latest alert data must be present if candid is None")
         candid = int(latest_alert_data["candidate"]["candid"])
 
     alert = None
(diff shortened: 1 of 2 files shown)


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -245,18 +257,26 @@ def post_alert(
...
         candids = {a.get("candid", None) for a in alert_data["prv_candidates"]}
         if latest_alert_data["candidate"]["candid"] not in candids:
             alert_data["prv_candidates"].append(latest_alert_data["candidate"])
 
-    df = pd.DataFrame.from_records(alert_data["prv_candidates"])
-    mask_candid = df["candid"] == str(candid)
-
-    if candid is None or sum(mask_candid) == 0:
+    if candid is None:
         candid = int(latest_alert_data["candidate"]["candid"])
```
guynir42, line 268: what if `latest_alert_data` is an empty dict here?

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -567,7 +589,7 @@ async def get(self, object_id: str = None):
                           data:
                             type: array
                             items:
-                              type: object
+                               type: object
```
guynir42, line 592: This looks like an accidental space (there should be two spaces)

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -647,6 +669,7 @@ async def get(self, object_id: str = None):
 
             if response.get("status", "error") == "success":
                 alert_data = response.get("data")
+                print(alert_data)
```
guynir42, line 672:
```suggestion
```

guynir42 (approved):
Some minor comments. Thanks for fixing this up!

extensions/skyportal/skyportal/handlers/api/alert.py
mcoughlin: Good call.

extensions/skyportal/skyportal/handlers/api/alert.py
mcoughlin: Removed.

extensions/skyportal/skyportal/handlers/api/alert.py
mcoughlin: Good idea, adding an extra check there.

PR #495, opened 2022-12-02
recompute padding fill value after normalization

nabeelre:
When padding a cutout that has been normalized with its median, the median must be recomputed after the normalization.


guynir42 (approved):
Good!

PR #502, opened 2023-02-09
Annotation button plugin to fetch/post SCoPe features from Kowalski/gloria

bfhealy:
This PR adds an annotation button plugin to fetch SCoPe-generated features from the `ZTF_source_features_DR5` catalog on gloria. The code performs a 2-arcsec cone search around the source's RA and Dec before querying gloria for a set of features. These features, which are then posted to the source as annotations, consist of all the light curve's 'basic stats' along with the ID, field, ccd and quadrant. The current code runs successfully but has the following limitations:

- The single-button nature of the frontend interaction does not permit the user to specify the feature catalog, cone search radius, or specific features to get
- A source can have multiple associated light curves, but since SCoPe features are generated on a per-lightcurve basis, the code currently only gets features from the first `_id` returned from the cone search.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/archive.py
@@ -291,7 +291,7 @@ class CrossMatchHandler(BaseHandler):
             return self.error(response.get("message"))
 
 
-class FeaturesHandler(BaseHandler):
+class ScopeFeaturesHandler(BaseHandler):
     @auth_or_token
     def post(self):
         """
--- a/extensions/skyportal/skyportal/handlers/api/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/archive.py
@@ -434,18 +434,18 @@ class FeaturesHandler(BaseHandler):
                     "query": {
                         "catalog": catalog,
                         "filter": filter,
-                    },
-                    "$project": {
-                        "_id": 1,
-                        "ra": 1,
-                        "dec": 1,
-                        "filter": 1,
-                        "meanmag": 1,
-                        "vonneumannratio": 1,
-                        "refchi": 1,
-                        "refmag": 1,
-                        "refmagerr": 1,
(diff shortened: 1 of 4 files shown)


extensions/skyportal/skyportal/handlers/api/archive.py
```diff
@@ -285,6 +291,251 @@ def get(self):
...
+                                    value = int(value)
+                                annotation_data[key] = value
+
+                        if annotation_data:
+                            annotation = Annotation(
+                                data=annotation_data,
+                                obj_id=obj_id,
+                                origin=catalog,
+                                author=author,
+                                groups=groups,
```
guynir42, line 515: Are you sure the right thing is to post different annotations for each feature? Wouldn't it be better to post one annotation with multiple keys?

guynir42: If the annotations all have the same origin I'm not sure you'd be able to post all of them (the origin+obj_id should be unique).

guynir42 (commented):
Looks fine. I think it is ok to have the front end use only the default values. If it matters to someone down the line and they want to use different input parameters we can always add a form or a choice in the profile or something. 

I think you should double check what happens when you try to post multiple annotations with the same origin.

extensions/skyportal/skyportal/handlers/api/archive.py
bfhealy: Thanks for the review @guynir42 - the code currently posts a single annotation containing all the different feature keys (see below). If I click the button a second time, it gives the expected 'Annotation already posted' message. 

Screen Shot 2023-02-13 at 10 59 14 AM

guynir42: OK I must have misread the loop and conditional indents! Looks good to me.

guynir42 (approved):

PR #503, opened 2023-02-10
Grab all thumbnails

mcoughlin:
This PR extends thumbnail endpoint to grab all for an object.


guynir42 (approved):
This will be super useful!

PR #505, opened 2023-02-17
catalogNames.filter type safety

Theodlz:
Hi reviewers. I haven't had the chance to run this code yet but I believe this could help prevent the source page crash in production. Please try running it before merging. I will try to do so myself and leave an "edit" here.


mcoughlin (approved):
Will check on preview.

PR #508, opened 2023-03-10
Use archive.js action instead of source.js for Scope features button

bfhealy:
This PR changes the source of `fetchScopeFeatures` from `source.js` (used during local testing) to `archive.js`. This will enable Scope features to be successfully fetched when clicking the new annotation button.


bfhealy:
@mcoughlin This change should enable scope features to be fetched on Fritz - kindly requesting your review.

mcoughlin (approved):
LGTM

PR #515, opened 2023-05-06
Alert query for multiple object IDs

mcoughlin:
This PR enables multiple object name alert queries.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -541,7 +541,7 @@ def get_alerts_by_id(
     projection : List[str]
         List of fields to query from database
     include_all_fields : bool
-        Include all fields in database. Defaults to False.
+        Include all available fields in database. Defaults to False.
     candid : Int
         Candidate ID to query
     """
@@ -595,7 +595,7 @@ def get_alerts_by_ids(
     projection : List[str]
         List of fields to query from database
     include_all_fields : bool
-        Include all fields in database. Defaults to False.
+        Include all available fields in database. Defaults to False.
     """
 
     # otherwise, run a find query with the specified filter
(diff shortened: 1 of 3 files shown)


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -493,12 +517,209 @@ def post_alert(
...
+    include_all_fields=False,
+    object_ids=None,
+):
+    """Get alert from database (by object ID).
+    ra : float
+        Right Ascension
+    dec : float
+        Declination
+    radius : float
+        Radius to search
```
guynir42, line 650: Specify the units (in the case of radius, say that units can be selected separately but specify what is the default).

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -493,12 +517,209 @@ def post_alert(
...
+    radius_units : str
+        Units of radius variable. Can be deg, arcmin, or arcsec.
+    program_id_selector : List[Int]
+        List of Program IDs to query
+    projection : List[str]
+        List of fields to query from database
+    include_all_fields : bool
+        Include all fields in database. Defaults to False.
```
guynir42, line 658: I'm not sure what "all fields" means.

guynir42 (approved):
Minor comments on the docstrings but otherwise looks fine.

guynir42 (approved):
Minor comments on the docstrings but otherwise looks fine.

extensions/skyportal/skyportal/handlers/api/alert.py
mcoughlin: Good point.

extensions/skyportal/skyportal/handlers/api/alert.py
mcoughlin: Yes.

PR #517, opened 2023-05-08
remove engine args

Theodlz:
As suggested by @stefanv, we shouldn't need to specify database arguments allowing more connections.


mcoughlin (approved):
LGTM

PR #518, opened 2023-05-09
More connections allowed in pool (db engine args)

Theodlz:
As suggested by @profjsb, we never use more than 1/3 of the max nb of connections to the database. Here, we try to allow more connections at once to prevent the frontend apps from filling up the pool and timing out too quickly.

### DIFF
--- a/fritz.defaults.yaml
+++ b/fritz.defaults.yaml
@@ -416,6 +416,7 @@ skyportal:
     port: 5432
     user: skyportal
     password: password
+    engine_args: {max_overflow: 30, pool_size: 20}
 
   ports:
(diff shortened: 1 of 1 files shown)


Theodlz:
@profjsb is this the kind of numbers you had in mind? Default is 15 and 10, here I used 30 and 20.

profjsb (approved):

mcoughlin:
@Theodlz do we want to merge this?

Theodlz:
@mcoughlin we can close it for now. It didn't create any issues recently, after improving the frontend to remove redundant calls.

PR #521, opened 2023-05-16
TNS info endpoint: wrap in try except

Theodlz:
We got some down times in prod when Kowalski returns something that is unexpected.


mcoughlin (approved):
LGTM

PR #522, opened 2023-05-26
Remove fritz-specific TNS interactions

mcoughlin:
This PR removes fritz-specific TNS interactions.


Theodlz (approved):
I never ever say no to removing code. LGTM

PR #523, opened 2023-06-08
Penquins 2.3.1 + SkyPortal 893af747e1b4290e23266b093f883a0d343cf122

Theodlz:
Needs https://github.com/skyportal/skyportal/pull/4349 to be merged first.
I suggest really trying out the code for this one, there's a lot that could break :)

Locally, I tried each handler and everything seems to work just fine, I can:
- Get alerts from the alerts page
- Visualize individual alerts and save them as as sources from the alerts page
- Get lightcurves from the persistent sources page, where we can now query gloria + melman (meaning DR5 but also now DR16)
- Save lightcurves as sources
- From the source page, get crossmatches with gloria + melman and see them on the centroid plot.
- From the source page, request scope features to be posted as annotations (from ztf features DR5 for now, but can be easily extended to any catalogs with DR`X` in a following PR (or migrated to ztf features DR16 once Brian uploads it to melman or gloria).

### DIFF
--- a/.requirements/ext.txt
+++ b/.requirements/ext.txt
@@ -1,2 +1,2 @@
-penquins>=2.1.0
+penquins>=2.3.1
 pymongo>=3.10.1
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
LGTM. I have also tested this locally and both alerts and light curve pathway seems to work as before.

PR #524, opened 2023-06-09
Persistent sources and Scope features interaction: Improvements

Theodlz:
- hide catalogs that do not start with ZTF_sources_202 (it's enough to cover the 20's)
- Scope annotation import: Allow user to pick scope catalog (DR3, DR5, and soon DR16)
- Archive.py: Allow to still query available instances of Kowalski when some of them aren't accessible, instead of blocking any access.


mcoughlin (approved):
Yes this is a better scheme.

PR #525, opened 2023-06-10
Alert/Persistent search ra/dec in hhmmss/ddmmss

Theodlz:
Search ZTF alerts and persistent sources using `RA/Dec` in `hh:mm:ss/dd:mm:ss` and `HHhMMmSSs/DDdHHhSSs` formats


mcoughlin (approved):
LGTM. It's a good add.

PR #526, opened 2023-06-10
Alert/Persistent search: radius units

Theodlz:
This PR allows users to use different units when specifying a radius for a conesearch on the Alerts and Persistent sources pages.


mcoughlin (approved):
Very convenient.

PR #527, opened 2023-06-27
Save ZTF alert photometry to other (nearby) sources

Theodlz:
- when posting an alert as a source, allow to post the data not only to the associated ZTF source but to any source within 2 arcsec. This will post photometry, but not the thumbnails as the other source could be an alert from WINTER, TURBO, PGIR,... which would replace its thumbnails. We could come up with a better behavior later.
- Thumbnail bugfix: There was a missing f in front of an f-string, rip...
- Sometimes a candid in prv_candidates doesn't exist in the alerts table. When that happens, we use the latest candid instead when posting thumbnails + notify the user.

I would suggest the reviewer to **really test this thoroughly**, as we wouldn't want existing features (see an alert and save it as a source) to not work anymore.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -1024,8 +1024,6 @@ class AlertHandler(BaseHandler):
         """
 
         data = self.get_json()
-        print(objectId)
-        print(data)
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
Tested on a number of sources, and behavior seems normal. Thanks @Theodlz.

PR #533, opened 2023-08-02
make image processing match existing method elsewhere

nabeelre:
The image processing currently here does not match what is being done on kowalski and what is expected by braii. This PR makes the image preprocessing consistent across the different systems.
discussed with @Theodlz


mcoughlin:
@nabeelre @Theodlz Is this still relevant?

nabeelre:
Yes, still relevant. I don't know why checks are failing because the logs aren't visible but the code should run fine

mcoughlin:
Ok @Theodlz you should weigh in on this then

PR #535, opened 2023-10-03
Temporary bugfix: prevent alert page from crashing when aux data is missing

Theodlz:
We are currently reingesting alert data in Kowalski that has been lost. 

In production, this caused issues when users try to access the ztf alert page for a given alert, if its aux data (aggregate of the prv_candidates + cross matches) hasn't been recovered yet.

This PR does:
- if aux data is missing, append the full alert history to effectively give the user all the detections (only the non-detections are now missing).
- if the aux data is missing, show a yellow warning on the frontend, which on hover explains what process we are going through right now, and what data is still available.


mcoughlin (approved):
Thanks for figuring this out Theo.

PR #536, opened 2023-10-05
Kowalski Filters - Autofollowup: set target group ids

Theodlz:
To prevent sending multiple triggers, and in general, to make sure that follow-up requests are shared with who should see them, we make it possible from Fritz's filter page to specify target_group_ids to use (on top of the filter's group id which is always there) when sending the request.

This PR has an associated Kowalski PR here: 
https://github.com/skyportal/kowalski/pull/255


mcoughlin (approved):
Yay frontend stuff!

Theodlz:
> Yay frontend stuff!

The frontend PR which essentially just adds a list to pick items from ends up being longer than the actual Kowalski PR that actually does something of it.

And then they ask people why they don't like React 👀

PR #540, opened 2024-02-15
ZTF Alerts: Forced Photometry

Theodlz:
* fetch the fp_hists from the aux data
* make the vega plot responsive and show the forced photometry on it
* better caching of the alert data (per objectid), just like we do for the source page
* rearrange the alert page using MUI Grids
* add toggles under the plot to show/hide the upper limits and the forced photometry
* add number input to dynamically change the SNR threshold applied to the forced photometry (to show as detections or upper limits).

This should provide all the tools one needs to see the alert data in Kowalski including the forced photometry with dynamic SNR cuts.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -4,7 +4,10 @@ import io
 import json
 import pathlib
 import traceback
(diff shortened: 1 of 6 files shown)


mcoughlin (approved):
LGTM

PR #541, opened 2024-03-16
Alerts Page: bulk save alerts & URL search params for both Alerts and Archive Pages

Theodlz:
* URL search params to pre-set the forms and run API query for both alerts and archival lightcurves.
* group alerts by objectID on the alerts page, to essentially get a list of objects, great for readability
* when grouping by objectID, make rows selectable to allow bulk saving of alerts as sources to a user's groups.
* cleanup the Archive page code

This PR requires https://github.com/skyportal/skyportal/pull/4830 to be merged first (once it's ready). The idea is that if one adds sources from another survey in Fritz, it's super useful to be able to grab all ZTF alerts (within a small enough radius) in Kowalski and save them, which thanks to https://github.com/skyportal/skyportal/pull/4830 allows a user to show the photometry of nearby sources on the photometry plot of any source.

TODO:
* limit the max radius that can be used when saving sources, to maybe something like 4 arcsec?
* make the pop-up window where a user saves alert(s) as source(s) smarter and more intuitive, showing info like "this source has already been saved to one or more groups".

### DIFF
--- a/extensions/skyportal/static/js/components/Alerts.jsx
+++ b/extensions/skyportal/static/js/components/Alerts.jsx
@@ -1,4 +1,5 @@
-import React, {useState} from "react";
+import React, { useState, useEffect } from "react";
+import { useSearchParams } from "react-router-dom";
 
 import Card from "@mui/material/Card";
 import CardActions from "@mui/material/CardActions";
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
Looks really nice

PR #549, opened 2024-05-25
Kowalski filters: support new keywords

Theodlz:
Add support for:
- autosave.saver_id, allowing Kowalski to use a different user account to autosave candidates as sources to a group, useful to not have everything posted by the `kowalski-bot` account both for book-keeping and to allow TNS autoreports from properly configured bot accounts.
- auto_followup.priority_order, so Kowalski knows if the instrument API associated to the allocation selected has descending or ascending priority values.
- auto_followup.implements_update, so Kowalski knows if the priority of an existing request can be updated or not.


mcoughlin (approved):
LGTM

PR #550, opened 2024-06-01
Frontend routes and imports 

Theodlz:
Adopts the new routes scheme where our frontend components are now nested in subdirectories in the static folder.


Theodlz:
self merging, as the docker build tests happens only on main anyway

PR #551, opened 2024-06-02
Docs GA

Theodlz:
An attempt at fixing the GA that generates the docs. Currently we have a couple of issues getting the python env to work

### DIFF
--- a/.github/workflows/build-test-doc_upload.yaml
+++ b/.github/workflows/build-test-doc_upload.yaml
@@ -25,6 +25,7 @@ jobs:
 
     - name: Install system dependencies
       run: |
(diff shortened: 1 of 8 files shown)


Theodlz:
pausing this PR while I update a bunch of SkyPortal dependencies to avoid conflicts with those pinned in Kowalski

Theodlz:
in the mean time (so I can keep trying to fix the rest), not referencing ALL the Kowalski dependencies in Fritz's requirements.txt

Theodlz:
also started to fix the integration tests but those need more work, however the main goal of this PR (fix the docs) is achieved and this can be merged

mcoughlin (approved):
LGTM

PR #553, opened 2024-06-19
ZTF Alert Post: stream data access issue

Theodlz:
I noticed that the `add_external_photometry()` method that the Fritz-specific AlertHandler POST method relied on (through the `post_alert()` function) treated group_ids and stream_ids as something to enforce on all datapoints (which to be fair is exactly how it is used everywhere else in the app!).

So here we simply batch photometry per stream and call `add_external_photometry()` on each batch to make sure that the stream data access is correct on these datapoints.


mcoughlin (approved):
LGTM

PR #555, opened 2024-07-10
WIP: negative fluxes when importing ZTF alert photometry

Theodlz:
At least now things won't be as confusing when adding data from an alert page vs when Kowalski sends the data to Fritz.

Next up is figuring out how we display the negative fluxes because everything in SkyPortal is designed to show negative flux measurements as non-detections, and its not so straightforward to make that customizable on a user basis.


mcoughlin (approved):
This looks great.

PR #556, opened 2024-08-05
make lightcurve_ids list if needed

mcoughlin:
This PR checks if lightcurve_ids is an integer and therefore need to be a list.


Theodlz (approved):
LGTM

PR #557, opened 2024-08-07
Docker compose update: migrate to new container name syntax

Theodlz:
docker-compose is now called with `docker compose` (space instead of a dash), and the the containers are not called `skyportal_web_1`, but `skyportal-web-1` (dashes instead of underscores).

This PR makes these changes + pins SkyPortal latest where these changes were made to the CI and `docker-compose.yaml` config too.


mcoughlin (approved):
LGTM

PR #558, opened 2024-08-21
Kowalski filter auto followup: ignore_allocation_ids

Theodlz:
Support for Kowalski+SP latest addition, that when editing a filter (with automated triggers turned on) lets a user set a list of allocations, that if any pending/completed requests exist for any of them, a new request's submission can be canceled.


mcoughlin (approved):
LGTM

PR #560, opened 2024-10-23
Documentation -> server list and endpoint summary

Theodlz:
This PR:
- adds fritz.science and preview.fritz.science as the list of servers to include in the auto-generated openapi specification.
- adds a `summary` to all of the endpoints Fritz adds on top of SkyPortal. This is required if we want to have a nice text rather than the raw URL shown in the list of endpoints on the left of the API documentation page.


Theodlz:
this PR is quite simple, self merging

PR #561, opened 2024-10-29
Kowalski Filter - not_if_tns_reported

Theodlz:
Add support for a new SkyPortal follow-up constraint (https://github.com/skyportal/skyportal/pull/5317) that can be used with Kowalski's auto triggering feature (https://github.com/skyportal/kowalski/pull/444).

This PR adds the API code to support it and validate its value, and adds the frontend required to edit it.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/kowalski_filter.py
+++ b/extensions/skyportal/skyportal/handlers/api/kowalski_filter.py
@@ -352,11 +352,30 @@ class KowalskiFilterHandler(BaseHandler):
                 "radius",
                 "implements_update",
                 "ignore_allocation_ids",
+                "not_if_tns_reported",
             }
             if not set(auto_followup.keys()).issubset(valid_keys):
                 return self.error(
                     f"auto_followup dict keys must be a subset of {valid_keys}"
                 )
+            if "not_if_tns_reported" in auto_followup:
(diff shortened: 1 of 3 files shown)


extensions/skyportal/skyportal/handlers/api/kowalski_filter.py
```diff
@@ -352,11 +352,30 @@ def patch(self, filter_id):
...
                 return self.error(
                     f"auto_followup dict keys must be a subset of {valid_keys}"
                 )
+            if "not_if_tns_reported" in auto_followup:
+                # if it's not empty or None, verify it's a a float or an int
+                if (
+                    auto_followup["not_if_tns_reported"] is not None
+                    and str(auto_followup["not_if_tns_reported"]).strip() != ""
+                ):
+                    try:
```
mcoughlin, line 367: Maybe also worth making note of the expected units here.

mcoughlin (approved):
Just a note to also add a # with the units if not already present.

extensions/skyportal/skyportal/handlers/api/kowalski_filter.py
Theodlz: Good point, I'll add that in before merging, thanks!

PR #563, opened 2025-04-28
alert.py: minor improvements

Theodlz:
In this PR we:
- add a decorator on top of each kowalski-related endpoints in `alert.py` which checks if kowalski is available. If it isn't then it tries to connect again before proceeding with the API call itself
- wrapped the Kowalski instantiation in a method that can be called by the decorator
- refactored the ztf alert cutouts import in a separate method
- cleaned up some of the code here and there, avoiding unnecessary operations and writes to disk when creating thumbnails

The original plan was to improve on performance but first it would be good to make the code a bit more robust to avoid running in some errors and such. There is still more work needed but this is a good first step.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -5,15 +5,15 @@ import json
 import pathlib
 import traceback
 from datetime import datetime, timedelta
-import arrow
 
-from astropy.time import Time
+import arrow
 import bson.json_util as bj
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import sqlalchemy as sa
 from astropy.io import fits
+from astropy.time import Time
 from astropy.visualization import (
     AsinhStretch,
     AsymmetricPercentileInterval,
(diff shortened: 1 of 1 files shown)


extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -33,41 +34,61 @@
...
-    Filter,
-    Classification,
 )
+from ...utils.parse import str_to_bool
 from ..base import BaseHandler
 from .photometry import add_external_photometry
 from .thumbnail import post_thumbnail
 
-alert_available = True
+matplotlib.pyplot.switch_backend("Agg")
```
mcoughlin, line 52: Should this be OS specific?

extensions/skyportal/skyportal/handlers/api/alert.py
```diff
@@ -830,6 +847,7 @@ def get_alerts_by_position(
 
 class AlertHandler(BaseHandler):
     @auth_or_token
+    @kowalski_available
```
mcoughlin, line 850: This is good tech.

mcoughlin (approved):
Nice

extensions/skyportal/skyportal/handlers/api/alert.py
Theodlz: It's required on some platforms I believe (if you run matplotlib in some backend) so I just like to add it by default. If your on a platform that needs it it's necessary anyways if you want it to run without warnings or issues, and on platforms that don't need it it won't change much of anything if you have that option on.

PR #564, opened 2025-05-05
Alert POST: thumbnails error bugfix

Theodlz:
We have some issues in prod introduced by #563 though it's not clear what the problem is exactly, but when calling the `post_thumbnail` method from SP we get an error telling us the data isn't valid image data. Not sure if this is because of the matplotlib backend change (that should also be valid on Linux) or because the format isn't specified to be `png` when calling the `savefig` method (which it was before, but from my local testing that wasn't required, perhaps that on other systems the default isn't png), but these changes seem to fix it in production (I added a hotfix in there).

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/alert.py
+++ b/extensions/skyportal/skyportal/handlers/api/alert.py
@@ -3,6 +3,7 @@ import gzip
 import io
 import json
 import pathlib
+import platform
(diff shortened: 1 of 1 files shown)


Theodlz:
In prod, with this hotfix, thumbnails are appearing as expected. merging

PR #578, opened 2026-05-21
Migrate Frontend to BOOM-based endpoints

Theodlz:
This PR adds all the endpoints and frontend changes necessary to power the alert search/display/import features (and other stuff like centroid plugins, etc) with BOOM instead of Kowalski.

The existing routes for Kowalski are left as is for backwards compatibility, but we should encourage folks to use the new BOOM routes. Once we know the Kowalski routes aren't used by anyone, we can retire them.


Theodlz:
@thomasculino is there a blocker to merge this? Would be good to update the skyportal version one more time before merging and get the ball rolling. Kowalski won't last long 😉

thomasculino (approved):
LGTM!

PR #588, opened 2026-06-09
Fix Boom tests

thomasculino:
The skyportal submodule bump pulled in the RTK Query migration, which removed fetchSource, checkSource, and fetchSources as dispatchable thunks from ducks/source.ts / ducks/sources.ts. The BOOM Alert.tsx and SaveAlertButton.tsx still imported and called them, so the Alert component crashed on mount with a TypeError — rendering nothing, causing all frontend test assertions to time out.

Fix: Replace the three removed thunk imports with their RTK Query equivalents (useLazyGetSourceQuery, useLazyCheckSourceQuery, useLazyFetchSourcesQuery) and update the call sites in Alert.tsx and SaveAlertButton.tsx to use the lazy hook trigger pattern instead of dispatch.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -111,6 +111,11 @@ jobs:
                 - /app/kafka_consumer
                 - ztf
                 - "20240617"
+            consumer-ztf-partnership:
+               environment:
+                 BOOM_API__AUTH__SECRET_KEY: ${BOOM_API__AUTH__SECRET_KEY}
+                 BOOM_API__AUTH__ADMIN_PASSWORD: ${BOOM_API__AUTH__ADMIN_PASSWORD}
+                 BOOM_KAFKA__CONSUMER__ZTF__SERVER: broker:29092
             scheduler-ztf:
               environment:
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Merge on green!

thomasculino:
I'm confused as to why the tests are not passing. Do you have any idea @mcoughlin?

mcoughlin:
@thomasculino I think we should merge RTK Query and then fix these.

thomasculino:
@mcoughlin You mean deploying Fritz with the RTK Query PR? If so, I'm updating files from the extensions folder that were modified by that PR and then we can go ahead and deploy

mcoughlin:
No I mean merging https://github.com/skyportal/skyportal/pull/6209 (Antoine is working through it) and then fixing this.

thomasculino:
I see. Do we know when it will be merged? I'd like to deploy fritz soon. There's too many changes between skyportal and production right now. I couldn't deploy end of last week and I've been trying to fix what was broken in the extensions folder since then.

mcoughlin:
@antoine-le-calloch is actively working on this stuff, I am sure as quick as is reasonable.

thomasculino:
Great, thanks!

PR #589, opened 2026-06-12
Fix BOOM alert pages crashing under RTK Query (builds on #588)

mcoughlin:
Builds on #588 (thomasculino:fix-test-frontend-boom) — branched from its current tip and layered the alert-component fix below, so it can be iterated to green independently.

## Why the boom frontend tests still fail on #588

skyportal's RTK Query migration (#6209) removed the `groups`, `source`, and `sources` redux slices. The alert components still read them via `useAppSelector((state) => state.groups.userAccessible / state.source / state.sources.latest)`, which now returns `undefined` → `TypeError` on render → the alerts **list and detail pages render blank**, so the boom frontend tests time out.

#588 converted the dispatchable thunks (`fetchSource`/`checkSource`/`fetchSources`) to lazy-query triggers, but left these **selectors** pointing at the removed slices. This reads them from the RTK Query hooks instead, matching skyportal's own migrated pattern:

- `groups` → `useGetGroupsQuery().data?.{userAccessible,all} ?? []`
- `source` → the `useLazyGetSourceQuery()` result's `.data`
- `sources` → the `useLazyFetchSourcesQuery()` result's `.data`

Files: `Alert.tsx`, `Alerts.tsx`, `SaveAlertButton.tsx`.

## Follow-ups (same root cause, not in the boom-test path, out of scope here)
`Archive.tsx` (removed `checkSource` thunk + `state.groups`), the filter plugins (`state.groups`/`state.profile`), and `CopyAlertPhotometryDialog` (`state.groups`) read the same removed slices and will need the same treatment.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -111,6 +111,11 @@ jobs:
                 - /app/kafka_consumer
                 - ztf
                 - "20240617"
+            consumer-ztf-partnership:
+               environment:
+                 BOOM_API__AUTH__SECRET_KEY: ${BOOM_API__AUTH__SECRET_KEY}
+                 BOOM_API__AUTH__ADMIN_PASSWORD: ${BOOM_API__AUTH__ADMIN_PASSWORD}
+                 BOOM_KAFKA__CONSUMER__ZTF__SERVER: broker:29092
             scheduler-ztf:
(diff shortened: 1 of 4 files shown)


thomasculino (approved):
All the CI is green, I must be dreaming

PR #594, opened 2026-06-14
async

mcoughlin:
This PR makes everything async.

### DIFF
--- a/extensions/skyportal/skyportal/handlers/api/kowalski/archive.py
+++ b/extensions/skyportal/skyportal/handlers/api/kowalski/archive.py
@@ -521,18 +521,12 @@ class ScopeFeaturesHandler(BaseHandler):
                         f'Cannot find source with id "{obj_id}". ', status=403
                     )
 
+                # Annotation is shared with the user's accessible groups.
                 # accessible_groups is lazy/sync under async; Group.select(user)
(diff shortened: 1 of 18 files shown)


extensions/skyportal/skyportal/handlers/api/kowalski/archive.py
```diff
@@ -498,19 +510,23 @@ def post(self):
...
-                groups = session.scalars(
-                    Group.select(session.user_or_token).where(Group.id.in_(group_ids))
+                # accessible_groups is lazy/sync under async; Group.select(user)
+                # is the async access-controlled equivalent (same set of groups).
+                groups = (
+                    await session.scalars(Group.select(session.user_or_token))
                 ).all()
+                group_ids = [g.id for g in groups]
 
                 if {g.id for g in groups} != set(group_ids):
```
thomasculino: Not new but isn't this dead code, since `group_ids` is built from `groups` as well? The condition will never be met

thomasculino (approved):
Let's point to skyportal main now that the other async PR has been merged, then we can merge

extensions/skyportal/skyportal/handlers/api/kowalski/archive.py
mcoughlin: Yup good point, deleted.

PR #596, opened 2026-06-19
Fix issues to deploy frontend

thomasculino:
Reverting last 3 commits and adding necessary fixes to deploy frontend. Later we will add back the commits that are reverted here


mcoughlin (approved):
LGTM. Hopefully it works! :)

PR #597, opened 2026-06-19
Migrate remaining legacy groups/filter/… reads to RTK Query

thomasculino:
Some files weren't properly updated, now everything should be working finally


mcoughlin (approved):
LGTM!

PR #598, opened 2026-06-21
Skip GKE auth/publish/deploy steps on forks

mcoughlin:
This PR skips the GKE auth/publish/deploy steps on forks.


thomasculino (approved):
Will be helpful!

PR #604, opened 2026-06-30
Parse numbers inside $in operator

thomasculino:
The $in operator was always parsing values to string which was a blocker for non string values such as fields.


mcoughlin (approved):
LGTM!

PR #583, opened 2026-05-31
test skyportal-plugin

mcoughlin:
This PR adds tests for the skyportal plugin.

PR #605, opened 2026-06-30
Add reverted commits async

thomasculino:
Add back the 3 reverted commits (#594, #583, #585).

Remove temporary fix by pinning skyportal to latest.

Align updates from the async PR to the skyportal updates. `db_stats.py` and `filter.py` were both modified on skyportal and Fritz. However, the updates on Fritz used a different syntax as the one used in Skyportal. The idea is to keep consistency between repositories.


mcoughlin (approved):
AWAIT WE GO!

PR #617, opened 2026-07-14
remove api.ts

mcoughlin:
This PR removes api.ts


extensions/skyportal/static/js/components/filter/kowalski/KowalskiFilterPlugins.tsx
```diff
@@ -464,84 +465,50 @@ const KowalskiFilterPlugins = ({ group }: KowalskiFilterPluginsProps) => {
...
-      if (json.status === "success") {
-        await dispatch(filterVersionActions.fetchFilterVersion(fid));
-      }
-    };
-
-    if (loadedId !== fid) {
-      fetchFilterVersion();
-    }
-  }, [fid, loadedId, dispatch]);
-
```
thomasculino, line 496: was that dead code?

thomasculino (commented):
Just one comment regarding deleted code. Otherwise looks good

extensions/skyportal/static/js/components/filter/kowalski/KowalskiFilterPlugins.tsx
mcoughlin: yeah rtk stuff takes care of it now.

PR #625, opened 2026-07-17
Fritz as true fork of skyportal

leonitousconforti:
## What this PR does

This PR converts fritz from a submodule-plus-overlay repo into a true git fork of skyportal, with a single merge commit (`a1289f61`) joining skyportal's full history (4,969 commits, through `d9e4b514`) into fritz's existing history. Nothing in fritz's history is rewritten every existing SHA, branch, issue reference, and PR link remains valid, and existing clones just `git pull`.

The resulting tree is exactly what `fritz build` produces today, baked in as source:

- skyportal at `d9e4b514` the same commit the submodule currently pins (important!!), so we start fully caught up with upstream
- the 113 files from `extensions/skyportal/` placed at their real skyportal paths (102 new files, 11 replacing skyportal counterparts), byte-identical to the overlay
- the `package.json` (7 JS deps) and `pyproject.toml` (4 python deps) merges that `launcher/skyportal.py:patch()` did at build time
- the `skyportal` submodule, `.gitmodules`, and `extensions/` retired; fritz's docs moved to `doc/fritz/`

## Why

Today the overlay is applied with a blind `copy_tree` at build time. Git knows nothing about the relationship between our copies and upstream's files, so every upstream sync means bumping the submodule and hoping our 11 overridden files haven't drifted. Divergence requires a lot of manual effort and checking

As a fork, upstream tracking becomes ordinary git:

```bash
git fetch upstream            # upstream = github.com/skyportal/skyportal
git merge upstream/main
```

## ⚠️ Merge requirement: merge commit only

**This PR **MUST** be merged with "Create a merge commit" not squash nor rebase

The entire value of this PR is the ancestry it introduces. Squashing 4,969 commits into one, or rebasing would rewrite them. Either way the shared history with the upstream is destroyed and `git merge upstream/main` becomes impossible again with "refusing to merge unrelated histories". 

## Alternatives explored

1. **Rewrite fritz's history with `git-filter-repo`** remapping `extensions/` throughout history, dropping the submodule, and stitching skyportal on top. It gives the cleanest per-file lineage however every fritz commit gets a new SHA, so it requires a force-push to `main`, lifting branch protection, every clone being reset, and it can't even be reviewed as a PR (GitHub refuses to compare unrelated histories). Rejected as far too disruptive. The trade off accepted instead is that fritz's old commits keep referring to `extensions/skyportal/...` paths. Tools like `git log --follow` and `git blame` should still work though with their builtin rename detection and given that the files are byte identical at their new paths.

2. **Fresh GitHub fork of skyportal.** Native fork badge and Sync button but repo's issues, discussions, and identity would be lost :(

## Known follow-ups

- `launcher/skyportal.py:patch()` and the copy/merge machinery are now dead code to be removed. `fritz build` can be simplified.
- CI: this tree carries skyportal's `.github/` workflows; fritz's ci will need porting to the fork layout.
- `LICENSE.txt` is fritz's; skyportal's version carries an MIT attribution for the periodogram code that should be folded in.
- fritz's old `tools/` scripts weren't carried over


leonitousconforti:
You can view the branch at https://github.com/leonitousconforti/fritz-as-skyportal-fork/tree/main since the diff tab won't be useful.

I would of course love fritz to be a true fork on GitHub (with the fork badge and sync buttons) for simpler maintenance with one-click buttons on the GitHub interface, however, I believe that this is a happy starting place / middle ground.

This could also make getting some of these changes merged in upstream SkyPortal (if we wanted to do that in the future) easier maybe?

What are people's thoughts? Obviously this is a big change and might require changes to the release repo too but I haven't look yet. I don't want to be too disruptive to people's workflows / other work, so if this isn't the time or place to do this I hear ya

leonitousconforti:
Here is the git "topology" of this pr/branch

```
fritz history (1,110 commits, untouched) ──> 0c17423 ──┐   ← fritz-marshal:main today
                                                        ├──> a1289f61  ← main (the PR branch)
skyportal history (4,969 commits) ───────> d9e4b514 ───┘   ← skyportal main = current submodule pin
```

The top line is fritz's original, unrewritten history, so a1289f61 is a plain descendant of what's on `fritz-marshal:main` today. a1289f61 is a commit that has two parents thought (fun fact: apparently this is actually how merge commits work under the hood, they have two parent commit), one parent holds fritz history and the other parent holds skyportal's history. This allows merges, rebasing, and everything else from upstream skyportal as they do share a full history now while also preserving the history of the `fritz-marshal:main` branch

PR #637, opened 2026-08-06
filter interface removal

mcoughlin:
This PR continues to remove the BOOM specific filter interface.


thomasculino (changes requested):
When creating a brand new filter, the broker_id is null. So, without setting the default to boom, we will not see the UI and won't be able to build the filter on the frontend. We should add that in the FilterPlugins.tsx

mcoughlin:
@thomasculino Yup good point.

thomasculino (approved):

PR #638, opened 2026-08-06
remove kowalski alerts endpoint

mcoughlin:
This PR removes the Kowalski alerts endpoint.


mcoughlin:
@thomasculino can we try doing this and see what happens?

thomasculino (approved):
Sure, let's do it! We might want to go over the docs when we finish moving things over to clean things up.

PR #642, opened 2026-08-11
the end of an era

mcoughlin:
This PR removes Kowalski from fritz.


thomasculino:
Do we want to get rid of the kowalski filter page? The filters won't be active anymore but is that what we want to do? I'm not sure about the answer.

mcoughlin:
@thomasculino I *think* so since that means things don't fall over when we take kowalski offline. I have the filters saved and will share them with you on slack now.

extensions/skyportal/skyportal/handlers/api/filter.py
```diff
@@ -1,263 +0,0 @@
...
-                group_id=body.group_id,
-                broker_id=body.broker_id,
-                altdata=body.altdata,
-            )
-            session.add(fil)
-            await session.commit()
-            return self.success(data={"id": fil.id})
-
-    @permissions(["Upload data"])
-    async def patch(self, filter_id: int, *, body: FilterPatchBody = None):
```
thomasculino, line 140: We need to update the patch filter in skyportal to allow filter renaming. Otherwise, names will get out of sync between boom and skyportal db.

thomasculino (approved):
One comment but LGTM!

extensions/skyportal/skyportal/handlers/api/filter.py
mcoughlin: Yes I will get that in. Thanks Thomas!

PR #643, opened 2026-08-12
remove fritz filter tests

mcoughlin:
This PR removes the last filter tests and handlers.


fritz.defaults.yaml
```diff
@@ -15,7 +15,10 @@ skyportal:
     #
     # base64.b64encode(os.urandom(50)).decode('ascii')
     secret_key: abc01234
-    factory: skyportal.app_server_fritz.make_app_fritz
+    factory: skyportal.app_server.make_app
+
+    # Bounded so the pool cannot exhaust the database connections.
+    thread_pool_size: 8
```
thomasculino, line 21: where do we use that?

thomasculino (approved):
One minor questions on that new key in `fritz.defaults.yaml`. I don't see where we use that key. Do we need to remove it or add support for it?

fritz.defaults.yaml
mcoughlin: @thomasculino It's in this PR. https://github.com/skyportal/skyportal/pull/6520 you can review it and I can pin it here if you have time.

thomasculino (approved):
LGTM! skyportal#6520 has been approved as well

